PR #19, opened 2018-07-30
Handle circular genome mega-clusters

SJShaw:
Prevents potential entire-genome clusters being formed due to a gene/CDS crossing the record origin in a circular genome (e.g. in NZ_CP016793.1).

For now, this is solved by splitting any bridging features into two new features. A more thorough version will need a large rewrite of any section of code that uses a location start and end point.

Also adds a minor change to enforce clusters having a valid `contig_edge` member, though this can still be `True` in circular genomes until clusters can be formed across the bridge.

### DIFF
--- a/antismash/common/secmet/features/feature.py
+++ b/antismash/common/secmet/features/feature.py
@@ -9,7 +9,10 @@ from typing import Any, Dict, List, Optional, Tuple, Union
 from Bio.SeqFeature import SeqFeature, FeatureLocation, CompoundLocation
 from Bio.Seq import Seq
 
-from antismash.common.secmet.locations import convert_protein_position_to_dna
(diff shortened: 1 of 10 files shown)


antismash/common/secmet/features/feature.py
```diff
@@ -9,7 +9,10 @@
 from Bio.SeqFeature import SeqFeature, FeatureLocation, CompoundLocation
 from Bio.Seq import Seq
 
-from antismash.common.secmet.locations import convert_protein_position_to_dna
+from antismash.common.secmet.locations import (
+    convert_protein_position_to_dna,
+    location_bridges_origin
```
kblin: The implementation is only added in the next patch, does it make sense to reorder these?

PR #26, opened 2018-08-22
ori-split cluster / border / CDS / domain fixes

user1313:


antismash/common/secmet/record.py
```diff
@@ -89,6 +89,7 @@ def __setattr__(self, attr: str, value: Any) -> None:
         except AttributeError:
             raise AttributeError("Record does not support dynamically adding attributes")
 
+    @property
```
SJShaw: is_circular was a method. You were testing if the method is None, not the results of the method. To be consistent with other code in the codebase, the correct approach would be to call the method itself by adding the parentheses.

user1313: Understood, thanks.. Reverting. Is this method being used anywhere else in the code so far? I could not find it but perhaps i missed it?

SJShaw: This one in particular isn't, as you have changed the lines that used it, but the intended style is that anything with a verb in the name is a function (e.g. `has_complement(), is_contained_by(), overlaps_with()`) or a method, while simple nouns would be properties or plain members (e.g. `.cds_children, .location, .results`).

antismash/common/secmet/features/feature.py
```diff
@@ -79,7 +76,8 @@ def get_sub_location_from_protein_coordinates(self, start: int, end: int) -> Fea
             return FeatureLocation(dna_start, dna_end, self.location.strand)
 
         new_locations = []
-        for location in sorted(self.location.parts, key=lambda x: x.start):
+        # do not sort here either way, as we want biological order. instead, reverse the location parts if needed
+        for location in reversed(self.location.parts) if self.location.strand == -1 else self.location.parts:
```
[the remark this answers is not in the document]
nestorzaburannyi: > This is only relevant for circular genomes, but you're applying it to all CDS features, even those not in a circular genome.

Trans-splicing? Theoretically possible and there is no harm in sorting biologically, or is it?

nestorzaburannyi:
i will rebase once we resolved all issues here and once the merge is possible

PR #33, opened 2018-09-27
Misc small fixes and improvements

SJShaw:
- adds `common.subprocessing.run_diamond()` and moves clusterblast to using it
- adds a function on Regions to fetch all unique grandchild clusters (`.get_unique_clusters()`)
- fixes some `tool` and `aSTool` behaviour in `secmet` features
- shifts critical logging to stderr instead of stdout
- some tweaks for style and to the pylint config (since many classes contain more than 4 unrelated pieces of data, max-args was too low)
- removes 3 unused profiles from hmm_detection and adds a test to ensure no more appear
- improves clusterblast commandline help
- makes genefinding prerequisite checking more thorough under `--check-prereqs`

and last, but most fun:
- left and right arrow keys now move to previous and next region in HTML output


antismash/detection/genefinding/__init__.py
```diff
@@ -49,7 +49,9 @@ def check_prereqs() -> List[str]:
     if options.genefinding_tool in ['none']:
         return failure_messages
     binaries = []  # type: List[str]
-    if options.genefinding_tool in ['prodigal', 'prodigal-m']:
+    if options.check_prereqs_only:
+        binaries = ["prodigal", "glimmerhmm"]
```
kblin: hmm, not sure about this one. if you never run on fungi, you never need glimmerh,,

kblin: otoh, you can just add a dummy script in that case

PR #37, opened 2018-10-03
RiPP cleavage fixes

SJShaw:
The 'start' position of a cleavage site was calculated and never used. Since all affected RiPP analyses used the beginning of the CDS translation as their leader start point anyway, the 'start' has been removed from all these functions and classes.

Also ensures that those regex-based cleavage site locators that subtract from the cleavage site position don't end up with a non-existant leader.


kblin:
Not sure what's up with codecov, according to buildbot the upload worked fine.

PR #39, opened 2018-10-05
technical changes in cluster/supercluster/region extension/merging logic + new BiosynML code

nestorzaburannyi:
and also minor fixes

### DIFF
--- a/antismash/detection/nrps_pks_domains/domain_identification.py
+++ b/antismash/detection/nrps_pks_domains/domain_identification.py
@@ -5,6 +5,7 @@
     features.
 """
 
+from collections import defaultdict
 import logging
 from typing import Any, Dict, List, Optional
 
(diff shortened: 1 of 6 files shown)


kblin:
Hi Nestor,

I think it'd be better if you could split this PR into three separate PRs. As far as I can see this PR has three separate parts that I feel have different likelihood to be accepted to master before we branch off the 5-0-stable release branch.

Realistically speaking, the circular genome rework is too big a change for me to be comfortable to merge it this short before the release. So I won't be merging a PR that contains this work before our 5-0-stable branch is cut, probably at some point in November.

There's the BiosynML part of this PR, which has a small enough impact that I'd be willing to merge it provided we can get some tests for it and ideally a spec for the file format so we can extend it when we add new features to antiSMASH. Without a clear idea how to extend BiosynML for new features, I'm worried that this output format will just slowly bit rot again just as it happened after the old version was added to version 3.

Last but not least there seem to be a bunch of bug fixes for issues you have identified in our master branch. Those I definitely want to see merged.

nestorzaburannyi:
Hi Kai,

You can close this now, as the automatic merge is not possible anymore again. Will do in smaller batches.

PR #43, opened 2018-10-10
nrps_pks: remove assert on comma

nestorzaburannyi:
fails on N-(1,1-dimethyl-1-allyl)Trp which is in antismash/modules/nrps_pks/external/NRPSPredictor2/data/labeled_sig

### DIFF
--- a/antismash/modules/nrps_pks/nrps_predictor.py
+++ b/antismash/modules/nrps_pks/nrps_predictor.py
@@ -40,7 +40,6 @@ class PredictorSVMResult(Prediction):
         self.single_amino_pred = str(single_amino_pred)
         assert ',' not in self.single_amino_pred
         self.stachelhaus_code = str(stachelhaus_code)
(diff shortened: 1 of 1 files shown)


kblin:
Are we sure this doesn't break downstream code that assumes that the only predictions that can contain `,` are the large and small group predictions?

nestorzaburannyi:
Kai, to whom is this question directed? No idea... I think you know your code much better than i do. If you have an idea on how to fix that issue in a better way, suggest, please. At the moment the code and the database source contradict each other.

SJShaw:
This was intended to catch the case that the predictor returned multiple options, but missed that multiple options in the stach codes were joined by `/` instead, unlike the other prediction results. The specific offender in the `N-(1,1-dimethyl-1-allyl)Trp` is also going to break the NORINE compatibility we were aiming for. 

I agree that the assert is currently wrong, but I think we need something else as a replacement rather than just removal. I'm not quite sure what though.

PR #45, opened 2018-10-11
secmet: biological exon order

nestorzaburannyi:


antismash/common/secmet/test/test_locations.py
```diff
@@ -60,9 +60,9 @@ def test_position_conversion_nonzero_compound(self):
         location = CompoundLocation([FeatureLocation(6, 15, strand=-1),
                                      FeatureLocation(21, 27, strand=-1)])
         assert len(location) == 15
-        assert self.func(0, 2, location) == (21, 27)
-        assert self.func(1, 4, location) == (9, 24)
-        assert self.func(3, 5, location) == (6, 12)
+        assert self.func(0, 2, location) == (9, 15)
```
SJShaw: That the order of start and end flips is concerning. I would expect that for any given ordered location with a consistent strand, one of `start < end` or `end < start` would apply to all combinations of valid protein locations.

antismash/common/secmet/locations.py
```diff
@@ -30,45 +30,46 @@ def convert_protein_position_to_dna(start: int, end: int, location: FeatureLocat
...
-            raise ValueError(("Converted coordinates %d..%d "
-                              "out of bounds for location %s") % (dna_start, dna_end, location))
-        return dna_start, dna_end
+    dna_start = start * 3
+    dna_end = end * 3
 
-    parts = sorted(location.parts, key=lambda x: x.start)
-    gap = 0
-    last_end = parts[0].start
+    processed = 0
```
SJShaw: Having been reminded of the existence of frameshifts causing overlapping location parts [here](https://github.com/antismash/antismash/pull/48/commits/74b18dd89047887e3cba68979e0416b672682c37), I don't think this method is going to work. Using the lengths of parts in this way is going to cause some problems because the counts ignore that the overlap is there.

GCF_000265205.3 is an example assembly that contains this kind of location.

nestorzaburannyi: exactly what i meant by adding the todo: in line 71 in the first of the two commits. Such a combined subroutine would only work on locations, processed = 0 will still exist there, but the locations will be returned, not dna_start, dna_end.

antismash/common/secmet/test/test_locations.py
nestorzaburannyi: again, i agree. if we implement the combined subroutine, this will not be the case anymore. Those tests will be changed to test locations against locations.

nestorzaburannyi:
since this is incompatible with the merged implementation of codon_start, this pull request has been replaced with  #50

PR #46, opened 2018-10-11
nrps_pks: return missing A-domain predictions to GenBank output

nestorzaburannyi:


nestorzaburannyi:
In modules/nrps_pks/results.py you have the following keys ever occurring in domain.predictions:

PKS signature
Minowa
KR activity
KR stereochemistry
consensus

Which one is the "tool making the prediction"? I would rather call them prediction methods, and so are the ones that i am trying to put back. Maybe you think that they were clobbering the GenBank, but i have reasons to thing that users would be happy to see the raw signals provided by all those methods. And since the generate_nrps_consensus() subroutine will simply return an "nrp" in case there is a tie, that information will be lost from GenBank. This change is so ridiculously small that i have no idea why we are even talking about it. Unless you can tell me, please, what is technically wrong with the code here, i do not see the reason why you requested changes here instead of closing the pull request.

SJShaw:
Those are all methods, albeit some with terrible names.  Minowa is the primary author of the paper used as the basis of the minowaCAL predictor. PKS_signature is the in-house name for a signature based PKS method, etc. All of the fields you're adding are related information. If it predicts a single amino, it will also have predicted all of the more general fields which include the single amino result, so there's no real reason to include most of the prediction details. There's also the case where you're adding these predictions without the caveat that NRPSPredictor has noted that the result is out of the applicability domain.

As there is no other A domain predictor in antiSMASH 5 right now (sandpuma is likely to return with a new version sometime next year), there's no ties to be found for A domains, a tie in the consensus call can't happen.  

The reason I didn't close it is that I feel like allowing a discussion is the best way to go, it also allows people to weigh in, as I'm not the only person involved in the project.  Since you feel I should have just closed it originally, I'll close it now.

PR #47, opened 2018-10-12
all_orf: change naming scheme of allorfs

nestorzaburannyi:
changed the naming scheme for allorfs as the current one fails in case two of the detection modules accept orfs with same counters (e.g. allorf004). We suggest using allorf_X_Y, where X is start and Y is end of this potential CDS candidate. Because you are checking if any new potential CDS overlaps any of the existing (or accepted by other modules, hence also existing, here i am not sure yet?), this should guarantee that there is no name collision.
Also removed the counter and made the label mandatory instead.

### DIFF
--- a/antismash/common/all_orfs.py
+++ b/antismash/common/all_orfs.py
@@ -89,26 +89,24 @@ def scan_orfs(seq: str, direction: int, offset: int = 0) -> List[FeatureLocation
     return sorted(matches, key=lambda x: min(x.start, x.end))
 
 
-def create_feature_from_location(record: Record, location: FeatureLocation,
-                                 counter: int = 1, label: Optional[str] = None) -> CDSFeature:
+def create_feature_from_location(record: Record, location: FeatureLocation, name: str) -> CDSFeature:
     """ Creates a CDS feature covering the provided location.
 
         Arguments:
             record: The Record the CDSFeature will belong to, used to generate
                     the feature translation
             location: The FeatureLocation specifying the location of the CDSFeature
-            counter: An integer to use to format a default label 'allorf' with,
-                     used only if label not provided
-            label: The locus tag, protein id, and gene name to use for the new
-                   CDSFeature
+            name: The locus tag, protein id, and gene name to use for the new
+                  CDSFeature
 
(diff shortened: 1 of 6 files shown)


antismash/common/all_orfs.py
```diff
@@ -89,26 +89,28 @@ def scan_orfs(seq: str, direction: int, offset: int = 0) -> List[FeatureLocation
...
-                   CDSFeature
+            name: The locus tag, protein id, and gene name to use for the new
+                  CDSFeature
 
         Returns:
             The CDSFeature created.
     """
-    if label is None:
-        label = 'allorf%03d' % counter
+    if not isinstance(record, Record):
```
SJShaw: This isn't really needed, since we're using methods of the record in this function anyway. Type checking is more important where the value is stored and used at some arbitrary later point (like many of the property setters throughout secmet). Value checks are still important though.

This function doesn't add the CDS to the record anyway, so it's a little strange to mention that in the error message.

antismash/common/all_orfs.py
```diff
@@ -89,26 +89,28 @@ def scan_orfs(seq: str, direction: int, offset: int = 0) -> List[FeatureLocation
...
+        raise TypeError("Can not create a feature, no record to attach it to")
+    if not isinstance(location, FeatureLocation):
```
SJShaw: This one will be raised when attempting to create the feature a couple of lines down, so it's not really neccessary either. All Feature subclasses will detect invalid locations at construction time.

antismash/common/all_orfs.py
```diff
@@ -89,26 +89,28 @@ def scan_orfs(seq: str, direction: int, offset: int = 0) -> List[FeatureLocation
...
+        raise TypeError("Can not create a feature without location")
+    if not isinstance(name, str):
```
SJShaw: This one might be better as ensuring that the string isn't empty
```suggestion
    if not name or not isinstance(name, str):
        raise ValueError("name must be a non-empty string")
```

kblin:
Should be fixed by 90f82442c, thanks for reporting this.

PR #52, opened 2018-10-18
secmet: move pseudo tag to feature class and make sure it is None in biopython

nestorzaburannyi:
At the moment, only gene features can be pseudogenes. A simple generalization here to all features.

### DIFF
--- a/antismash/common/secmet/features/feature.py
+++ b/antismash/common/secmet/features/feature.py
@@ -51,19 +51,20 @@ class Feature:
         subclass.
     """
     __slots__ = ["location", "notes", "type", "_qualifiers", "created_by_antismash",
-                 "_original_codon_start"]
+                 "_original_codon_start", "_pseudo"]
 
     def __init__(self, location: FeatureLocation, feature_type: str,
                  created_by_antismash: bool = False) -> None:
         assert isinstance(location, (FeatureLocation, CompoundLocation)), type(location)
         if location_bridges_origin(location):
             raise ValueError("Features that bridge the record origin cannot be directly created: %s" % location)
+        self._pseudo = False # only ever set to True in from_biopython() when relevant
         assert location.start <= location.end, "Feature location invalid: %s" % location
         self.location = location
         self.notes = []  # type: List[str]
         assert feature_type
         self.type = str(feature_type)
-        self._qualifiers = OrderedDict()  # type: Dict[str, List[str]]
+        self._qualifiers = OrderedDict()  # type: Dict[str, Optional[List[str]]]
         self.created_by_antismash = bool(created_by_antismash)
         self._original_codon_start = None  # type: Optional[int]
 
@@ -178,6 +179,10 @@ class Feature:
             return self.location.start in other and end in other
         raise TypeError("Container must be a Feature or a FeatureLocation, not %s" % type(other))
 
+    def is_pseudo(self) -> bool:
+        """ Was the feature marked as a pseudo feature"""
+        return self._pseudo
+
     def to_biopython(self, qualifiers: Dict[str, Any] = None) -> List[SeqFeature]:
         """ Converts this feature into one or more SeqFeature instances.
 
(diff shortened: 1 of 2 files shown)


antismash/common/secmet/features/feature.py
```diff
@@ -201,9 +208,14 @@ def to_biopython(self, qualifiers: Dict[str, Any] = None) -> List[SeqFeature]:
             # adjust location back if neccessary
             if self._original_codon_start != 0:
                 feature.location = _adjust_location_by_offset(feature.location, -start)
+        if self._pseudo:
+            quals["pseudo"] = []
```
SJShaw: Why not just set the value to `None` straight away? It removes the need for a special case 3 lines later.

antismash/common/secmet/features/feature.py
```diff
@@ -51,13 +51,16 @@ class Feature:
...
-                 "_original_codon_start"]
+                 "_original_codon_start", "_pseudo"]
 
     def __init__(self, location: FeatureLocation, feature_type: str,
-                 created_by_antismash: bool = False) -> None:
+                 created_by_antismash: bool = False, pseudo: bool = False) -> None:
         assert isinstance(location, (FeatureLocation, CompoundLocation)), type(location)
         if location_bridges_origin(location):
             raise ValueError("Features that bridge the record origin cannot be directly created: %s" % location)
+        self._pseudo = bool(pseudo)
```
SJShaw: Since you're changing the code anyway, I'd rather see this as simply set to `False`, with no constructor argument for it, since construction with a truthy value isn't allowed. This would be an improvement over the original code.

antismash/common/secmet/features/feature.py
```diff
@@ -51,13 +51,16 @@ class Feature:
...
+        if self._pseudo:
+            assert not created_by_antismash, "pseudo %s features can only come from input files" % feature_type
```
nestorzaburannyi: where to put this assert then? Delete or move it to to_biopython()?

SJShaw: ```self._pseudo = False``` is required, because otherwise it can potentially not exist to check in `is_pseudo()`.

The rest can be deleted, since it can no longer be set to `True`.

antismash/common/secmet/features/feature.py
```diff
@@ -51,13 +51,14 @@ class Feature:
...
     __slots__ = ["location", "notes", "type", "_qualifiers", "created_by_antismash",
-                 "_original_codon_start"]
+                 "_original_codon_start", "_pseudo"]
 
     def __init__(self, location: FeatureLocation, feature_type: str,
                  created_by_antismash: bool = False) -> None:
         assert isinstance(location, (FeatureLocation, CompoundLocation)), type(location)
         if location_bridges_origin(location):
             raise ValueError("Features that bridge the record origin cannot be directly created: %s" % location)
+        self._pseudo = False # direct creation is only allowed with falsy value
```
SJShaw: Comments should typically indicate _why_ the code does something, rather than repeating what the code says.

```suggestion
        self._pseudo = False # only ever set to True in from_biopython() when relevant
```

nestorzaburannyi:
now there is a failed build:

antismash/common/secmet/features/feature.py:210: error: Incompatible types in assignment (expression has type "None", target has type "List[str]")

This is why i think i added those 3 additional lines in the first place...

nestorzaburannyi:
still get build errors:
antismash/common/secmet/features/feature.py:194: error: Unsupported left operand type for + ("None")
antismash/common/secmet/features/feature.py:194: note: Left operand is of type "Optional[List[str]]"

SJShaw:
> still get build errors:
> antismash/common/secmet/features/feature.py:194: error: Unsupported left operand type for + ("None")
> antismash/common/secmet/features/feature.py:194: note: Left operand is of type "Optional[List[str]]"

If you change a variables type and don't update the interactions with that variable, that will happen, yes.

The point of a code review is not to hand off half finished things to someone else and have them do all the remaining work. If you like, I can make the full changes myself, but it will be when I have time.

nestorzaburannyi:
I think the code was finished work before you requested changes. Shall i revert to the original commit?

SJShaw:
No. I'll do it myself.

nestorzaburannyi:
Thanks.

PR #53, opened 2018-10-18
secmet: change empty "/sec_met: Type:" to "/sec_met: Type: unknown", if no core domains are associated with CDS

nestorzaburannyi:
unspecified would be probably better, but to be consistent with the domain specificity predictions, keep it "unknown"?


SJShaw:
No longer relevant as of PR #62

PR #54, opened 2018-10-18
secmet: get x in "/sec_met: Type: x" tags from the core domains annotated on that CDS

nestorzaburannyi:
instead of cluster type that overlap CDS. For example, a pure nrps CDS was annotated as nrps-t1pks if an t1pks clusters overlapped it. See NC_021658, location 3653023-3667491


SJShaw:
No longer relevant as of PR #62

PR #55, opened 2018-10-22
 cds_feature: make sure translation fits the feature

nestorzaburannyi:
record NZ_QUMU01000001 and many more from RefSeq are affected

### DIFF
--- a/antismash/common/secmet/features/cds_feature.py
+++ b/antismash/common/secmet/features/cds_feature.py
@@ -161,6 +161,16 @@ class CDSFeature(Feature):
             logging.warning("Translation for CDS %s (at %s) has a gap. Discarding and regenerating.",
                             locus_tag or protein_id or gene, bio_feature.location)
             translation = ""
+        # ensure translation is not longer than the feature
+        elif translation and len(translation) > len(bio_feature.location) // 3:
+            logging.warning("Translation for CDS %s (at %s) is longer than expected. Discarding and regenerating.",
+                            locus_tag or protein_id or gene, bio_feature.location)
+            translation = ""
+        # ensure translation is not shorter than the feature, allow for 1 aa difference (stop codon)
+        elif translation and len(translation) < len(bio_feature.location) // 3 - 1:
+            logging.warning("Translation for CDS %s (at %s) is shorter than expected. Discarding and regenerating.",
+                            locus_tag or protein_id or gene, bio_feature.location)
+            translation = ""
         if not translation:
             if not record:
                 raise ValueError("no translation in CDS and no record to generate it with")


kblin:
Can't reproduce with sequence downloaded from https://www.ncbi.nlm.nih.gov/nuccore/NZ_QUMU01000001 today, the only warning I get is `WARNING  23/10 11:28:26   Fasta header too long: renamed "NZ_QUMU01000001.1" to "NZ_QUMU01000001"`, which is expected.

nestorzaburannyi:
Kai, can you please try the whole genome instead NZ_QUMU01000000 ? I have also uploaded an improved commit.

kblin:
So what ID do I run? I'm trying to get some other work done as well.

kblin:
Also, I'd argue that of the record provides translations, we should trust them.

nestorzaburannyi:
> So what ID do I run? I'm trying to get some other work done as well.
Sorry for your time wasted, i think i've nailed it to at least NZ_QUMU01000015 from that genome.

> Also, I'd argue that of the record provides translations, we should trust them.
In ideal World, yes. Please consider:

- User excising parts of the genome.
- User changing some CDS location by drag&and&drop.
- Currently fails even on some RefSeq entries (mostly draft genomes, like the one above). Don't you think our aim should be at least 100% NCBI/MiBIG compatibility?

I hope this is pretty much simple test that should catch any strange translation. Anyway, if it is too long and there are domains/motifs on the overhang part, Antismash will crash. If it is too short (like, user was extending CDS to an upstream start codon, but did not care about updating /translation), the annotated stuff might be a complete nonsense anyway.

kblin:
>     * User excising parts of the genome.

That would make the leftover parts pretty meaningless, wouldn't it?

>     * User changing some CDS location by drag&and&drop.

Sounds like the tool supporting that drag&drop operation should fix the translation after changing the location. I mean, if the drag&drop doesn't change the feature size, we'll still work on a wrong translation. So basically you'd argue that because we can't trust the translations, we should always re-generate them. But then we'll get translations wrong where there's non-canonical codon use, especially as by a similar step in logic we shouldn't trust the translation table specified in the file.

>     * Currently fails even on some RefSeq entries (mostly draft genomes, like the one above). Don't you think our aim should be at least 100% NCBI/MiBIG compatibility?

The proper solution here is to fix the NCBI annotation by reporting this error to the NCBI.

> I hope this is pretty much simple test that should catch any strange translation. Anyway, if it is too long and there are domains/motifs on the overhang part, Antismash will crash. If it is too short (like, user was extending CDS to an upstream start codon, but did not care about updating /translation), the annotated stuff might be a complete nonsense anyway.

Again, by the same argument we can't ever trust the translations because the user might have moved a feature around without changing the size.

The current solution is easy to code and easy to communicate: Garbage in, garbage out.

nestorzaburannyi:
> > ```
> > * User excising parts of the genome.
> > ```
> 
> That would make the leftover parts pretty meaningless, wouldn't it?
> 
> > ```
> > * User changing some CDS location by drag&and&drop.
> > ```
> 
> Sounds like the tool supporting that drag&drop operation should fix the translation after changing the location. I mean, if the drag&drop doesn't change the feature size, we'll still work on a wrong translation. So basically you'd argue that because we can't trust the translations, we should always re-generate them. But then we'll get translations wrong where there's non-canonical codon use, especially as by a similar step in logic we shouldn't trust the translation table specified in the file.

I get your point... but it might take ages for the tools to catch up...

> 
> > ```
> > * Currently fails even on some RefSeq entries (mostly draft genomes, like the one above). Don't you think our aim should be at least 100% NCBI/MiBIG compatibility?
> > ```
> 
> The proper solution here is to fix the NCBI annotation by reporting this error to the NCBI.
> 
I did. The response was roughly along the same lines what i usually get for my pull requests :)

> > I hope this is pretty much simple test that should catch any strange translation. Anyway, if it is too long and there are domains/motifs on the overhang part, Antismash will crash. If it is too short (like, user was extending CDS to an upstream start codon, but did not care about updating /translation), the annotated stuff might be a complete nonsense anyway.
> 
> Again, by the same argument we can't ever trust the translations because the user might have moved a feature around without changing the size.
> 
> The current solution is easy to code and easy to communicate: Garbage in, garbage out.

By this argument, we should start with creating the Universe... and then sequencing the bug again.
What about recycling the garbage a bit and getting the best of it? Never again?

PR #56, opened 2018-10-23
 feature: shrink CDS on minus strand if codon_start > 1, instead of expanding it

nestorzaburannyi:


SJShaw:
The offset you're trying to negate is already negated. The tests show this, and if you'd run them you'd have noticed.

Closing this as a non-issue.

nestorzaburannyi:
Thanks Simon, appreciated. What about NZ_QUMU01000042? May i kindly ask you to run it on the current master and comment why it fails on "assert cds_within_regions  # because every cluster should have genes'? I think internally, we have a very fishy "CDS(ATI61_RS49645, [<0:>4320](-))". Those fishy CDS (it seems that every -1 strand, with codon_start > 1 are affected) are most of the time hidden from our eyes, but in this case it fortunately surfaced.

With this patch is use, we will hit another bug down the line caused by the nearly flawless implementation of codon_start, but that's another story. The thing is, we need to adjust codon_start before we hit the part which is checking/regenerating translation. As i insisted multiple times, codon_start should be tied to the CDSFeature class, not Feature. What do you think?

Please consider re-opening PR #56 .

P.S. Tests are valuable, but only if they are written correctly in the first place. If you indeed agree with re-opening, i can remedy those tests providing you are not working on them as well.

kblin:
Hi Nestor,

maybe we need to tackle this differently. How about you open an issue for every issue you identify, explaining what issue you see with which record, like you did in the previous comment.

If you just open a pull request with a seemingly random change that breaks our tests and has zero explanations, we're spending time trying to understand why that change is necessary, and then might decide to close it like this one, which means that you'll have wasted your time creating the PR in the first place.

As we're all busy people, let's try to optimize this workflow.

Cheers,
Kai

nestorzaburannyi:
if Kai doesn't want to have garbage in -> at least some result out (#55), the other bug i mentioned is not then a bug and is irrelevant.

PR #58, opened 2018-10-24
Run full analyses in RiPP tests

SJShaw:
This PR isn't fixing anything, but is necessary for simplifying planned changes to the `sec_met` annotations and any other future detection annotation changes.

Previously the RiPP integration tests used detection annotations in genbanks to avoid failing when detection modules had issues. This causes problems when detection gets out of sync, with the genbanks needing to be regenerated each time detection changes.

This PR removes this shortcut of detection and runs the detection on each cluster. This does make the integration tests take a little longer, but on my machine all RiPP module tests (unit and integration) take a total of 85 seconds, which I think is still reasonable.

### DIFF
--- a/antismash/modules/lanthipeptides/test/integration_lanthipeptides.py
+++ b/antismash/modules/lanthipeptides/test/integration_lanthipeptides.py
@@ -4,12 +4,8 @@
 # for test files, silence irrelevant and noisy pylint warnings
 # pylint: disable=no-self-use,protected-access,missing-docstring
 
-import json
 import os
 import unittest
-
-from helperlibs.bio import seqio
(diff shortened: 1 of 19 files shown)


antismash/modules/lanthipeptides/test/integration_lanthipeptides.py
```diff
@@ -26,6 +21,7 @@ def setUp(self):
         self.options = build_config(["--minimal", "--enable-lanthipeptides"],
                                     isolated=True, modules=antismash.get_all_modules())
         self.set_fimo_enabled(True)
+        self.expected_html_content = None
```
kblin:
```suggestion
```
This seems unused.

PR #59, opened 2018-10-24
 prepeptide: fix string formatting

nestorzaburannyi:


SJShaw:
This was fixed in PR #70 as commit 49b1c05, along other instances

PR #79, opened 2018-12-13
More HTML fixes and tweaks

SJShaw:
Fixes the following:
- broken contact link
- clusterblast gene tooltips not appearing
- copy to clipboard functionality not working (the new version copies automatically, no ctrl-C required)
- broken MiBIG links in focus panel
- focus panel being shared for multiple regions

Also improves output style for various NRPS/PKS parts that didn't previously take advantage of expanders.


SJShaw:
Now also fixes some issues with AT signature classification generation, mostly about representation. PKS monomer consensus was also updated to use all classifications, not just the first of equals.

PR #85, opened 2019-01-02
secmet: Cope with more than gap characters in CDS translations

kblin:
Signed-off-by: Kai Blin <kblin@biosustain.dtu.dk>


kblin:
Implemented  by #87 already.

PR #88, opened 2019-01-03
Misc fixes and cleanup

SJShaw:
A collection of small fixes and cleanup changes, some of which have been gathering dust for a while.

Changes that aren't superficial/style:
- the JSON results object now stores a schema version, at some point we may be able to handle backwards compatibility but for now a mismatch raises an error
- the HMM rule parser no longer allows a single condiiton inside a cds condition e.g. `cds(AMP-binding)`
- taxon information isn't lost on reuse
- improved handling of some edge cases for inputs (protein records, empty record sets, etc)
- products and their matching rules in the HTML output now consistently pair correctly
- `Prepeptide` features are now properly annotated with the `tool` qualifier being `antismash`, matching all other features created by antiSMASH


antismash/outputs/html/__init__.py
```diff
@@ -60,19 +62,36 @@ def write(records: List[Record], results: List[Dict[str, ModuleResults]],
...
+def copy_template_dir(template: str, output_dir: str, pattern: Optional[str] = None) -> None:
     """ Copy files from a template directory to the output directory, removes
-        any existing directory first
+        any existing directory first. If pattern is supplied, only files within
+        the template directory that match the template will be copied.
+
+        Arguments:
+            template: the source directory
+            output_dir: the target directory
+            only_matching: a pattern to restrict to, if given
```
kblin: Shouldn't this be `pattern`?

SJShaw: It should be, yes. I just fixed that now.

PR #90, opened 2019-01-07
Input processing cleanup

SJShaw:
- adds a custom error class for simplified catching of bad inputs and uses it where relevant
- refactors some of the input parsing code that was more complex than it should have been
- fixes an issue that all CDS_motif features were being stripped, when only ones created previously by antiSMASH should have been stripped

### DIFF
--- a/antismash/common/errors.py
+++ b/antismash/common/errors.py
@@ -1,0 +1,19 @@
+# License: GNU Affero General Public License v3 or later
+# A copy of GNU AGPL v3 should have been included in this software package in LICENSE.txt.
+
+""" Contains various error classes for use throughout antiSMASH """
+
+
+class AntismashError(Exception):
+    """ A general catch-all for errors within antiSMASH """
+    pass
+
+
+class AntismashInputError(AntismashError):
+    """ An error for when input to antiSMASH as a whole are invalid or contain
+        unsupported content.
+
+        Not intended to replace TypeError/ValueError in functions not directly
+        parsing or converting input files.
+    """
+    pass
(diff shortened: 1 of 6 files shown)


antismash/common/record_processing.py
```diff
@@ -46,44 +45,34 @@ def parse_input_sequence(filename: str, taxon: str = "bacteria", minimum_length:
...
     except Exception as err:
-        logging.error('Parsing %r failed with unhandled exception: %s',
-                      filename, err)
+        logging.error('Parsing %r failed: %s', filename, err)
         raise AntismashInputError(str(err)) from err
 
+    for record in record_list:
+        if minimum_length < 1 \
+                or len(record.seq) >= minimum_length \
+                or 'contig' in record.annotations \
```
kblin: Does this make sense? We can't handle contig/WGS_SCAFLD/WGS records for now, and on the web service the downloader takes care of these.

antismash/common/errors.py
```diff
@@ -0,0 +1,20 @@
+# License: GNU Affero General Public License v3 or later
+# A copy of GNU AGPL v3 should have been included in this software package in LICENSE.txt.
+
+""" Contains various error classes for use throughout antiSMASH """
+
+from typing import List
```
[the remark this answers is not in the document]
SJShaw: Nope, it's a leftover from an experimental dependency error I was creating. Good catch, I'll remove it.

antismash/common/record_processing.py
SJShaw: The condition is unchanged from what it was previously, it will at a later stage (in `pre_process_sequences()`) raise an error on those cases. I'm happy enough to move that error to the parse step instead, if you'd prefer.

kblin: Nah, we probably just need to revisit this if/when we add a `ncbi-acc-download` dependency to deal with WGS records.

PR #92, opened 2019-01-09
Handle errors in clusterblast database

SJShaw:
Reference clusters in the clusterblast database (from `clusterblast_20170105_v8_31.tar.xz`) have a listing of locus tags and protein ids. For 28 of those clusters (of ~220,000), the database is missing details of one or more proteins (e.g. `DF820425` and protein `GAJ32366`, presumably due to sharing a locus tag with `GAJ32367`). This caused crashes when those particular clusters had hits *and* were in the top scores.

Until such point as the clusterblast database is regenerated and these issues are removed, this PR fixes the problem by removing the offending clusters, along with any proteins they contained. Any hits against those removed proteins will be ignored.

The runtime cost is around 2 seconds on my machine per load of the database. Running diamond on the database will take a few minutes, so this is manageable for the moment.

Also included is a workaround for a bad interaction between `minimock` and `unittest.mock`, which only appeared when running the test file directly (e.g. `pytest antismash/modules/clusterblast/test/test_clusterblast.py`).

### DIFF
--- a/antismash/modules/clusterblast/core.py
+++ b/antismash/modules/clusterblast/core.py
@@ -188,6 +188,33 @@ def load_reference_proteins(accessions: Set[str], searchtype: str) -> Dict[str, 
     return proteins
 
 
+def strip_clusters_missing_proteins(clusters: Dict[str, ReferenceCluster],
+                                    proteins: Dict[str, Protein]) -> None:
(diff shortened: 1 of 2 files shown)


kblin:
I think we can move our code over to use `unittest.mock` in the long run, now that that's in the stdlib.

PR #95, opened 2019-01-25
Another web interface update

SJShaw:
Changes most of the HTML output for modules, with the key features being:
- splitting module output into selectable tabs for body details and sidepanel details
- adds help tooltip functionality for sections (only NRPS/PKS domains has this as a demo, full help still needs to be written)
- revamps NRPS/PKS domain images completely, with new sizing, appearance, the ability to filter by selected features, and the ability to select features by clicking the feature name

The region selector buttons have changed due to observing user confusion with highly fragmented inputs. Collapser/expander icons have changed for the same reason.

A preview is available [here](https://antismash.secondarymetabolites.org/upload/s_collinus/index.html) (at least temporarily).


antismash/outputs/html/templates/overview.html
```diff
@@ -14,6 +14,7 @@
   <meta property="og:type" content="website" />
   <meta name="twitter:card" content="summary" />
   <meta name="twitter:site" content="@antismash_dev" />
+  <meta name="viewport" content="width=device-width, initial-scale=1">
```
kblin: My bad, this seems to make things worse on my phone.

kblin:
I'll go merge this anyway, we can always revert that one change. I guess the meta tag would be better if we had a mobile-optimised layout, but we really don't.

PR #103, opened 2019-02-05
new rule for C-nucleosides

marnixmedema:


antismash/detection/hmm_detection/data/truD.hmm
```diff
@@ -0,0 +1,496 @@
+HMMER3/f [3.1b2 | February 2015]
+NAME  truD_seed
```
kblin: This should probably be called `truD` just like in the cluster rules and the hmmdetails file.

marnixmedema: Changed now!

kblin:
Covered by #110

PR #106, opened 2019-02-06
- added a rule for RaS-RiPPs

marnixmedema:
See
https://pubs.acs.org/doi/10.1021/jacs.8b11060
and
https://pubs.acs.org/doi/full/10.1021/jacs.8b10266


kblin:
Covered by #110

PR #108, opened 2019-02-07
docker,clusterblast: handle docker container building without a mounted data dir

SJShaw:


SJShaw:
The automated tests won't check the Dockerfile changes, so it's probably worth double-checking with local checkout to be sure it's functioning as expected.

PR #109, opened 2019-02-07
Name clarity changes

SJShaw:
Cluster structures:
- cluster -> protocluster (with genbank names being `protocluster` and `proto_core`
- supercluster -> candidate cluster (with genbank being `cand_cluster` due to length limits)

Cluster types:
Changes product joiners to commas instead of hyphens, which allows the following
- `pksfragment` -> `pks-like`
- `nrpsfragment` -> `nrps-like`

Also changes the TTA codon legend button to plural form like the other legend buttons.


antismash/common/secmet/features/test/test_region.py
```diff
@@ -150,6 +150,7 @@ def test_genbank(self):
             region.write_to_genbank(output.name)
             bio = list(seqio.parse(output.name))
         assert len(bio) == 1
+        print(bio[0].features)
```
kblin: I guess this is a leftover from debugging.

kblin: Ah, this is a feature, I see. It's fine in that case.

PR #110, opened 2019-02-08
New rules

SJShaw:
Combines #103, #105, and #106:
- updates the tests that were being broken by the changes
- cleans up unnecessary commits
- ensures CSS rules will match their targets
- adds relevant information to rule comments where possible

Also improves the CSS component matching test to ensure case sensitivity of HTML classes is respected.


antismash/detection/hmm_detection/cluster_rules.txt
```diff
@@ -278,3 +278,12 @@ RULE RaS-RiPP
     CUTOFF 5
     EXTENT 10
     CONDITIONS PF04055 and TIGR01716
+
+RULE fungal-RiPP
```
kblin: While we now have a case sensitivity check, I see no added value in making this rule have mixed case.

PR #111, opened 2019-02-08
Rule labelling revamp

SJShaw:
- changes rules to be capitalised when relevant, e.g. `pks-like` -> `PKS-like` and `ppysks` -> `PpyS-KS`
- stops HTML generation from trying to title-case products, now that leading lower case can be meaningful, e.g. `hglE-KS`

For compatibility with existing results from the antiSMASH-db and elsewhere, links to documentation remain as lowercased versions.


antismash/modules/nrps_pks/orderfinder.py
```diff
@@ -34,8 +35,7 @@ def analyse_biosynthetic_order(nrps_pks_features: List[CDSFeature],
     compound_predictions = []  # type: List[SuperClusterPrediction]
     # Find NRPS/PKS gene superclusters
     superclusters = [cluster for cluster in record.get_superclusters()
-                             if "nrps" in cluster.get_product_string()
-                             or "pks" in cluster.get_product_string()]
+                             if will_handle(cluster.products)]
```
kblin: This is a functional change in the Balhimycin gene cluster, where we now no longer report monomers for T3PKS.

PR #121, opened 2019-03-18
Support partial transAT-PKS clusters

SJShaw:
antiSMASH 5 tightened the rules on what was considered a transAT-PKS cluster, but this lost the partials that were being predicted. We've had feedback that users still wanted those partial clusters detected, so this adds a new `transAT-PKS-like` cluster type, similar to our other partial PKS and NRPS clusters.

Other tiny tweaks:
- removes a demo comment line now that the first few rules contain an example of that
- changes the rule count test from a static number due to having to change it every time rules are updated


antismash/detection/hmm_detection/test/test_hmm_detection.py
```diff
@@ -158,7 +158,7 @@ def test_find_clusters(self):
 
     def test_create_rules(self):
         rules = hmm_detection.create_rules(self.rules_file, self.signature_names)
-        assert len(rules) == 52
+        assert len(rules) == open(self.rules_file).read().count("\nRULE")
```
kblin: Doesn't this somewhat defeat the "check we didn't accidentally delete a rule" character of the test?

kblin: I agree this test was quite cumbersome, so maybe we don't really want to use it that way.

PR #125, opened 2019-03-27
Rule strictness options

SJShaw:
Adds a strictness option to `hmm_detection` (`--hmmdetection-strictness` on the command line), with options of `strict`, `relaxed`, and `loose`.

`relaxed` is the default value and has the same behaviour as the old versions
`strict` removes the `*-like` variants from detection
`loose` adds the clusterfinder rules `cf_saccharide` and `cf_fatty_acid` (which have been renamed to remove the prefix) and the new `halogenated` rule (which could still use a good description on the docs site).

Also bundles in a small fix for when a region is created with only subregions.


antismash/detection/hmm_detection/cluster_rules/loose.txt
```diff
@@ -20,3 +20,9 @@ RULE fatty_acid
     CUTOFF 20
     NEIGHBOURHOOD 10
     CONDITIONS  bt1fas or ft1fas or t2fas or fabH
+
+RULE halogenated
+    COMMENT why
```
kblin: Halogenases are frequently involved in secondary metabolite biosynthesis.

SJShaw: Sounds good, I'll use a modified version of that for the docs description.

PR #132, opened 2019-04-12
Move CSS generation from lessc to PySCSS

SJShaw:
Removes tracking of CSS files built from source by lessc and instead regenerates them on demand using the python SCSS module. CSS files will now be rebuilt when source files are updated (or built files are missing), just like all the HMMer databases.


antismash/outputs/html/css/secmet.scss
```diff
@@ -0,0 +1,276 @@
+/* colour definitions for secondary metabolite types */
+
+@mixin secmet($fg: black, $bg: white, $highlight: orangered){
+    background-color: $bg;
+    a, a:link, a:visited {
+        color: $fg;
+    }
+    rect& {
```
SJShaw: I don't know why this line highlights red, but `rect&` generates correctly as `rect.NRPS` etc, where `rect.&` does not.

PR #133, opened 2019-04-16
Use a ClusterBlast database generated from the antiSMASH DB

kblin:
Also changes the format of the `known` and `sub` databases to enforce unique IDs across the board.


antismash/modules/clusterblast/core.py
```diff
@@ -150,11 +151,11 @@ def load_reference_clusters(searchtype: str) -> Dict[str, ReferenceCluster]:
     return clusters
 
 
-def load_reference_proteins(accessions: Set[str], searchtype: str) -> Dict[str, Protein]:
+def load_reference_proteins(names: Set[str], searchtype: str) -> Dict[str, Protein]:
```
SJShaw: `names` is now unused and can be removed

antismash/modules/clusterblast/core.py
```diff
@@ -233,12 +226,11 @@ def load_clusterblast_database(record: secmet.Record, searchtype: str = "cluster
                 a dictionary mapping cluster name to Cluster instance
                 a dictionary mapping protein name to Protein instance
     """
-    accessions = set()
+    names = set()
```
SJShaw: As `load_reference_proteins()` doesn't need this anymore, it's unused and can be removed

antismash/modules/clusterblast/core.py
```diff
@@ -781,29 +776,50 @@ def check_diamond_db_compatible(database_file: str) -> bool:
...
+        Returns:
+            The database version as an integer
+    """
     with open(database_file, 'rb') as handle:
         chunk = handle.read(16)
 
     if len(chunk) != 16:
         logging.debug("Database %s appears corrupted.", database_file)
-        return False
+        raise ValueError()
```
SJShaw: This should probably have a message so that if it escapes in future it'll be informative, also tests of this function could use it at some point

PR #139, opened 2019-04-21
clusterblast: diamond 0.9.24 finds more hits, adjust integration tests

kblin:
Signed-off-by: Kai Blin <kblin@biosustain.dtu.dk>


kblin:
Too fragile. It seems like #140 is the better solution

Issue #122, opened 2019-03-18
DeepBGC integration


prihoda:
Hi @kblin and team. We are happy about your involvement in the DeepBGC module, would you be interested in integrating the module in antiSMASH?

DeepBGC can be used programmatically, here's an example:

```python
from deepbgc import DeepBGCDetector, DeepBGCClassifier

os.environ["DEEPBGC_DOWNLOADS_DIR"] = os.path.join(os.path.basename(__file__), 'data')

detector = DeepBGCDetector('deepbgc', merge_max_protein_gap=1, merge_max_nucl_gap=2000, score_threshold=0.5)
product_classifier = DeepBGCClassifier('product_class')
activity_classifier = DeepBGCClassifier('product_activity')

detector.run(record)
product_classifier.run(record)
activity_classifier.run(record)
```

If you are interested, please let me know and I will be happy to get involved and discuss next steps.

kblin:
Hi David,

We've discussed this previously (I had a quick chat with @Microbiology about this after the preprint was released). I think conceptually DeepBGC complements the rule-based nicely. Practically, I still see a number of challenges with integrating DeepBGC with antiSMASH. I'm happy to discuss those, though.

So I think in a recent update, you changed DeepBGC to also accept pre-annotated genomes, that does remove one of the issues I had on my list, thanks. That said, in antiSMASH we also run our own PFAM annotations. Since running full genome PFAM is the largest chunk of the runtime of DeepBGC and takes around 2-3 orders of magnitude longer than a regular antiSMASH run without PFAM annotations, I would like to be able to at least re-use these annotations for both tools.

Another issue I see is from a packaging perspective with the tensorflow library. We have seen things break in fun and entertaining ways between minor versions of the library and are not particularly fond of adding it on top of the already quite significant dependency stack. I'm aware that by design this is not something you can easily fix in DeepBGC, so one thing we discussed internally was if we could come up with a way to load DeepBGC annotations into antiSMASH, and then have antiSMASH also display the DeepBGC regions in the output. We don't currently have the code to do this, but this would be trivial to implement.

I'm flexible as to how we connect things up from a UI perspective on the submission website. We could have a separate website for running DeepBGC that then allows passing on the results to antiSMASH, or support this directly from the antiSMASH submission page, and then internally have a processing step that first runs DeepBGC and then runs antiSMASH.

Does that sound like a solution you might be interested in?

prihoda:
Hi Kai, 

[...]

PR #163, opened 2019-05-22
External annotation sideloading

SJShaw:
Implements the proposed sideloading of external annotations as per #122.

**Input**
Input files are used by using the `--sideloader FILENAME` option. As each file can only have one tool specified, multiple files can be provided as `--sideloader FILENAME_A,FILENAME_B` to bundle multiple tools worth of annotations together into the antiSMASH result.

An example of a multi-record input with a variety of detail values is given in `antismash/detection/sideloader/test/data/good.json`. While only one each of subregion and protocluster is added in this example, as many as desired can be added.

The schema and subschemas are in `sideloader/schemas/general`. The loading functions can be used standalone (without the rest of antiSMASH) and are

**Output**
- Genbank: subregion and protocluster annotations are made with all provided area details, but only tool name, tool configuration is not currently included
- HTML: subregions and protoclusters are shown as per other areas of the same type, detail tabs are created for each area with all details of both the area and the tool configuration

**Potential issues:**
- sideloaded protoclusters will never quite function the same as detected protoclusters, without an additional section of specifying which CDS features are considered the defining CDS features (which then leads to a much more complex schema, as those CDS features need to have which profiles are mentioned, and so on)
- for sideloaded protoclusters to trigger specific analysis modules (like the t2pks module), the product names must match those in the `hmm_detection` rules exactly
- sideloaded subregions are never coloured, so telling multiple tools worth of inputs
- the `hmm_detection` module will always run, so multiple protocluster annotations of the same area can be created if the external annotations include a matching area
- protoclusters added that do not match an existing product will not have any colouring in the HTML output

The `antismash.js` used in this PR is also a prototype, before merging this would need to be updated to an actual release.


SJShaw:
It occurred to me that if we want to support multiple versions of the schema, we might want to embed a version number into the sideloaded JSON. If we only want to support the latest or at most a backwards compatible latest, then it may not be neccessary.

SJShaw:
Issues found using the DeepBGC example:
- subregion numeric identifiers in the region view reset to 1 in each JS-generated SVG, where the detail tabs keep the correct numbering
- regions with no protoclusters/`unknown` products don't have a contrast colour, so it's difficult to see by indicators which region is currently shown
- region labels overflow when the subregion is small, while it may cost a lot of vertical space, the labels could be move above similar to protoclusters
- detail tabs might be improved with a label for the distinct sections of tool information and area information

SJShaw:
Updated and known issues resolved.

An extra command line option has been added `--sideloader-simple`, which can specify a single accession and area which will be marked as a `manual` tool. This is simplified as it is largely intended for debugging and the like rather than for real tool annotations.

Also changed is that `hmm_detection` now marks all profile hits in full regions which aren't covered by protoclusters it finds.

kblin (commented):
Apart from this one JSON error, looks good.

kblin (approved):
Looks good to me 👍

PR #300, opened 2020-10-12
Update clusterblast dataset to antiSMASH-DB v3

SJShaw:
A slight change in database endpoints means a slight change in the FASTA headers, so older formats are now caught as part of the prerequisite checking step.


kblin (approved):
Looks good to me 👍

PR #311, opened 2020-10-28
Update mypy to a py3.8 compatible version

SJShaw:
mypy 0.630 is incompatible with python 3.8. Some extra type hints were required in order to update to mypy 0.790


kblin (approved):
Looks good to me 👍

PR #312, opened 2020-11-04
Add Pfam visualisation to HTML output

SJShaw:
Adds Pfam domain visualisation in a similar method to the existing NRPS/PKS domain visualisation.

image

Profile names that would be too long are replaced with `...`. Domains are coloured as per gene functions if there's a known association, the classifications may need to be refined.

Changed behaviour:
- overlapping Pfam domains are no longer reported, an overlap is resolved by (in order): normalised score by cutoff, length, start location, then name
- fixed `HmmerHit` end location being offset by one from the actual hit
- `HmmerResults` now regenerates exactly from the existing data as per other `ModuleResults`, with the filtering step based on current parameters now an explicit call in `regenerate_previous_results` for the relevant modules


kblin (approved):
Looks good to me 👍

PR #288, opened 2020-09-02
RREFinder

Alexamk:
Integration of precision mode of RREFinder into antiSMASH

PR #314, opened 2020-11-07
Change AntismashDomain to be more generic and subclassable by modules for specifics

SJShaw:
As discussed in #288, there needed to be a more generic way of handling `aSDomain` features annotated in GenBank files.

Subclasses of `AntismashDomain` can now be registered by modules creating them and registered as a variant for specific regeneration (via `antismash.common.secmet.features.antismash_domain.register_asdomain_variant()`).
After registration, `AntismashDomain.from_biopython()` will use the relevant type based on the `aSTool` qualifier matching the registered type. If no tool matches, the generic `AntismashDomain` will be used.

Previous NRPS/PKS-specific attributes are now handled by the `antismash.detection.nrps_pks_domains.ModularDomain` class.


kblin (approved):
Looks good to me 👍

PR #325, opened 2020-12-10
Cluster comparison module

SJShaw:
Adds a new cluster comparison module that takes into account multiple metrics for the comparison:
- identity, in a vaguely similar way to clusterblast
- components, comparing complete NRPS/PKS modules, gene functions, and secondary metabolite domains annotated by the detection modules
- order, taking into account the size of contiguous chunks, the gap length between the chunks, and whether chunks reverse strand differently to their neighbours

Analysis can be performed by searching for the query in the reference or searching for the reference in the query.
The areas to use for comparisons can be specified, comparing full reference region to full query region, or reference region to individual query protoclusters, or reference protoclusters to query protoclusters.

The analysis can be invoked via `--cc-mibig` for the MIBiG comparison. There is an antiSMASH database building right now that will be `--cc-asdb`. The option for supplying custom databases also exists with the `--cc-custom-dbs` option, providing a comma separated list of database description JSON that can specify the name, path, URL for matching regions, and which analysis modes and directionality to use.

A database build script is included to build a new database (but not the description and URL info file) from a directory of existing antiSMASH results (and optionally with MIBiG JSON data in the same directories).

Some underlying changes were required for full functionality. The largest of which was allowing for modules to generate custom datasets for the javascript portion of HTML output to access. There were also some miscellaneous serialisation and Genbank issues were found and fixed. Diamond database cleaning was generalised. The CDS `sec_met` qualifier can no longer able to be set to `None`, but can be empty. The diamond database check functionality was also moved from clusterblast to common, as it was able to be reused.

Likely future changes:
- replacing the overview HTML page with the clustercompare results instead of clusterblast's


antismash/modules/cluster_compare/results.py
```diff
@@ -0,0 +1,286 @@
+# License: GNU Affero General Public License v3 or later
+# A copy of GNU AGPL v3 should have been included in this software package in LICENSE.txt.
+
+""" Contains the results classes for the nrps_pks module """
```
kblin, line 4: Copy & pase error? :wink:

kblin (changes requested):
Apart from the two minor issues I commented on, it looks pretty good.

antismash/download_databases.py
```diff
@@ -307,20 +315,41 @@ def download_clusterblast(db_dir: str) -> None:
...
-        PFAM_LATEST_URL,
-        PFAM_LATEST_VERSION,
-        PFAM_LATEST_ARCHIVE_CHECKSUM,
-        PFAM_LATEST_CHECKSUM,
-    )
-
-    download_resfam(args.database_dir)
-
-    download_clusterblast(args.database_dir)
+#    download_pfam(
```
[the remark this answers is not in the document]
SJShaw: I absolutely did. Fixed with a rebase just now.

antismash/modules/cluster_compare/results.py
SJShaw: Also fixed with a rebase just now.

kblin (approved):
Looks good to me 👍

PR #327, opened 2020-12-14
html: update javascript to 0.8.1

SJShaw:
fixes an issue where the first detail tab is Pfam domains
and there's no clustercompare results for the region


kblin (approved):
Looks good to me 👍

PR #328, opened 2020-12-15
Misc fixes

SJShaw:
Major fixes:
 - a crash that could occur in `rrefinder` when multiple profiles hit the same CDS
 - detection rules with a `not` component could ignore that component under certain circumstances
 
Minor fixes:
- a typing issue in `hmm_rule_parser` that mypy doesn't detect
- `--sideloader-simple` help was incomplete
- `rrefinder` duplicated hits against CDS features when two or more relevant protoclusters included that CDS
- incorrect docstrings in `common.hmmer`


kblin (approved):
Looks good to me 👍

PR #331, opened 2020-12-17
html: update javascript to 0.8.2

SJShaw:
fixes an issue where swapping to and from a region with
clustercompare results could cause the comparison SVG
to disappear


kblin (approved):
Looks good to me 👍

PR #335, opened 2020-12-22
common.hmm_rule_parser: remove inferiors on overlap, not just contains

SJShaw:
Previously, nearly all rules with `SUPERIORS` specified were single CDS features and wouldn't create overlapping protoclusters. With the addition of the new RiPP rules, especially the `RRE-containing` rule, the previous method of filtering out 'inferior' types where the inferior was completely contained by the superior was insufficient and has been changed to be overlapping with any superior type.


kblin (approved):
Looks good to me 👍

PR #345, opened 2021-02-17
Detect multi-CDS NRPS/PKS modules

SJShaw:
Adds support for detecting valid complete modules that continue across CDS features.

Conditions for merging module fragments:
- both CDS features must be on the same strand
- CDS features involved must be sequential (features without NRPS/PKS domains aren't included in this check)
- the resulting module must be a valid complete module as per existing rules for modules

A preview from kirromycin with module lids:
image

A preview from kirromycin without module lids:
image


kblin (approved):
Looks good to me 👍

PR #355, opened 2021-04-13
Misc small additions and fixes

SJShaw:
The JSON output now includes a list of areas for each record, where previously they were derived from the other contents of the JSON. This should simplify things for people wanting something similar to the old spreadsheet style result and having to script it. 
This change is write-only so while the schema version is bumped (to 2), previous schema versions can still be reused with `--reuse-results`.

Other fixes:
- improves handling of some odd ways of specifying codon starts
- hides the pyScss Deprecation/FutureWarnings about nested sets, as users were thinking that was an error (can be removed when/if pyScss updates)
- changes the `ValueError` raised when `--genedfinding-tool` is set to `error` (the default), to a `AntismashInputError` so that the error message is displayed and the program exits cleanly (though still non-zero)

### DIFF
--- a/antismash/common/secmet/features/feature.py
+++ b/antismash/common/secmet/features/feature.py
@@ -292,7 +292,11 @@ class Feature:
         if leftovers:
             feature.created_by_antismash = leftovers.get("tool") == ["antismash"]
             if "codon_start" in leftovers:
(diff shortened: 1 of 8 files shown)


kblin (approved):
Looks good to me 👍

PR #388, opened 2021-10-04
clusterblast: adjust tests to handle newer diamond versions

SJShaw:
This is due to diamond 2 changing scoring results. Scores themselves seem to be only subtly changed, in the range of 0.1% from version 0.9, but this has resulted in more hits for clusterblast.

Tested with 2.0.9, specifically, as that's shipped with Debian Bullseye.


kblin (approved):
Looks good to me 👍

PR #389, opened 2021-10-12
Add support for dynamic profiles in cluster detection

SJShaw:
Not all types of hits necessary for rules are able to be found with HMMer. This adds support for code-based profiles that mimic HMMer hits for the purposes of cluster detection rules  in `hmm_detection`.

Any source file added to `antismash/detection/hmm_detection/dynamic_profiles` will be automatically searched for dynamic profiles (specifically, any variable that is an instance of `DynamicProfile` or a subclass of it) when antiSMASH is imported. Multiple profiles per file is acceptable, as long as each profile is a top level variable in the file.

Profiles duplicating a name of either another dynamic profile or an HMM in the signature list file cause an error to be raised, during the gathering step or when processing the signatures file for HMMs.

Two other small changes are included:
- unification of all imports of `HSP` from biopython to a single module within antiSMASH, since that import is accessing private submodules in `SearchIO`
- fixed `genefunctions` using `HMMResult` as an `HSP`-like replacement while ignoring the JSON conversion methods, this makes it easier to swap out with something like the newly added `ProfileHit`

**Example**
An example of a profile that arbitrarily creates hits when a locus tag matches `A`:
```python
from typing import Dict, List
from antismash.common.hmm_rule_parser.structures import DynamicHit, DynamicProfile
from antismash.common.secmet import Record

PROFILE_NAME = "locus_A_finder"

def find_hits(record: Record, target: str = "A") -> Dict[str, List[DynamicHit]]:
    hits = {}
    for cds in record.get_cds_features():
        if cds.get_name() == target:
            hits[cds.get_name()] = [DynamicHit(cds.get_name(), PROFILE_NAME)]
    return hits

# this variable will be detected as an instance of Dynamic profile and added dynamically
profile = DynamicProfile(
    PROFILE_NAME,
    "A profile for finding CDS features with the name 'A'",
    find_hits,
)
```

**Potential issues**
- `DynamicHit` instances can be given evalues and bitscores to match HMMer hits, these values (defaulting to 1.0 and 0.0, respectively) will be recorded in the domain qualifiers in the same way as HMMer hits. This could be misleading.
- Similarly, all `DynamicProfile` instances will have a 'seeds' value as per HMM profiles, which may be misleading when annotated as it is always zero.


kblin (approved):
Looks good to me 👍

kblin (approved):
17th time's the charm, I'm sure.

PR #398, opened 2021-10-28
chore: add a github action for comparing pull requests

SJShaw:
Add a github action that runs over pull requests and generates a report of resolved/added issues.


kblin (approved):
Looks good to me 👍

PR #399, opened 2021-11-04
Fix incorrect merging of cross-CDS NRPS/PKS modules

SJShaw:
It was possible for NRPS/PKS modules to be formed when the genes in question weren't sequential. This is now fixed.

This also incidentally fixes a crash in the bubbles visualiser that occurred when a module like this was generated and crossed candidate cluster boundaries,


kblin (approved):
Looks good to me 👍

kblin (approved):
Looks good to me 👍

PR #401, opened 2021-11-22
Misc fixes

SJShaw:
Fixes:
- domain docking analysis could split a CDS into parts if that CDS contained a cross-CDS module
- candidate polymers duplicated monomers from cross-CDS modules
- regions containing no protoclusters (e.g. those created by sideloaded subregions) would crash HTML generation
- the images on the overview HTML page weren't properly using product categories for colouring
- NRPS/PKS domain bubble view would misalign the SVG when the first domain was a modification domain
- PFam/TIGRFam visualisations would tile horizontally in 'expanded' mode when the domain names and genes were quite small

Improvements:
- docking-domain order analysis for PKS candidate clusters no longer counts CDS features with cross-CDS modules against the maximum feature limit


kblin (approved):
Looks good to me 👍

PR #403, opened 2021-11-23
Improvements to bubble view and cross-CDS modules

SJShaw:
Candidate clusters can contain only half of a cross-CDS module even when the CDS features are consecutive (e.g. NZ_CP020102.1 in the region 1,763,795 - 1,869,041). When drawing these kind of modules, it should a) not crash, and b) not show the module as complete when the other half is not included in that particular drawing subset.

This PR trims modules such as these down to the domains within the candidate and marks the resulting truncated module as incomplete for the purposes of drawing that particular candidate cluster result.


kblin (approved):
Looks good to me 👍

PR #405, opened 2021-12-10
Misc fixes

SJShaw:
Fixes colouring of region buttons in the HTML overview and in the record overview SVG either missing or defaulting to hybrid as a result of the category CSS changes.

Also fixes RiPP modules doubling up on adding new CDS features when protoclusters overlap (only seen so far with multiple lanthipeptide classes in the same region, but handled in all RiPP modules just in case).


kblin (approved):
Looks good to me 👍

PR #407, opened 2021-12-16
ClusterCompare MIBiG database update

SJShaw:
With the addition of new NRPS/PKS profiles and changes to how NRPS/PKS modules are detected, and many other changes, the accuracy of the clustercompare comparisons was decreasing.

This PR uses the current master of antiSMASH on MIBiG JSON as at `2.0-11abfec` to regenerate the comparison database.

Some fixes for the database generation are also included:
- gene function now only considers detection-stage gene function annotations
- subclasses of `Other` category MIBiG entries no longer build with mistaken comma joins for each character in the subclass
- the `--mibig` build option now takes a file with accessions to include, to more easily separate pending and active entries that may reside in the same input directory


kblin (approved):
Looks good to me 👍

PR #412, opened 2021-12-21
nrps_pks: Add NRPS C/E/Cy domain activity prediction

zreitz:
Members of the NRPS condensation domain superfamily (C, E, Cy) have well-conserved active sites that can be used to predict catalytic activity ([10.1016/j.bbapap.2017.05.010](https://doi.org/10.1016/j.bbapap.2017.05.010)).  This draft identifies and reports the activity and active site motif for each C/E/Cy domain.

C/E/Cy domains are aligned to a C domain superfamily alignment from Rausch et al. 2007 ([10.1186/1471-2148-7-78](https://doi.org/10.1186/1471-2148-7-78)) and the active site is extracted (of the form `HHIILDG`).  C and E domains have the canonical active site of `HHxxxDG`, while Cy domains have `DxxxxDx`.  The ~750 MIBiG clusters with C domains were used for testing the requirements. Relaxing the C/E site to `xHxxxxx` (the key catalytic residue) and Cy site to `Dxxxxxx` removed false "inactives" without introducing false "actives".  The remaining 88 "inactives" were manually checked and correspond well to domains known/suspected to be inactive in literature. 

The exception was a group of siderophore C domains from Actinobacteria that have a modified `HxxxxDx` motif, but are active based on the produced compounds. Allowing `HxxxxDx` as active would introduce false positives. The C domains form a single clade, so instead a new profile HMM called `Condensation_sid` was created from the domains (from BGC0000325, BGC0000349, BGC0000351, BGC0000423, BGC0001211, BGC0001752, BGC0001767). A bitscore cutoff of 375 was determined by searching against UniProt and MIBiG. `Condensation_sid` is otherwise treated as any other C domain in domain_drawing, domain_identification, and module_identification.

The activity and active site motifs are saved under `c_activity` and `c_activesite` in the NRPS/PKS results, and reported under "Substrate predictions". Additionally, inactive C domains are marked with an X in the Module tab (like KR domains).


kblin (approved):
Looks good to me 👍

PR #426, opened 2022-03-09
secmet: handle biopython introducing spaces into locus tags

SJShaw:
Long CDS names (e.g. AB366635.1 with a `gene` qualifier of `MIPLSYAQRRMWFINRFEGPSSTYNIPLLLRFHGSLDTAALRAAFQDVLV`) are line wrapped by biopython output when using a longer qualifier name, and when those are read back in with biopython it keeps a single space where the line break was.

This PR changes all `locus_tag` qualifier parsing in features that need to point at CDS features to remove any spaces found.


kblin (approved):
Looks good to me 👍

PR #434, opened 2022-03-31
switch from crc32 to md5 to properly hash locations without clashes

jlv100:
crc32 "is not cryptographically strong, and should not be used for authentication or digital signatures" according to [zlib docs](https://docs.python.org/3/library/zlib.html#zlib.crc32) .  

We ran into a clash rather quickly with the following locations, which result in the exact same crc32 digests:

    >>> from zlib import crc32
    >>> checksum=crc32('[5206081:5207221](-)'.encode('utf-8'))
    >>> f"{checksum:x}"
    '6dea7c77'
    >>> checksum=crc32('[1964876:1965035](-)'.encode('utf-8'))
    >>> f"{checksum:x}"
    '6dea7c77'

A hash isn't really necessary because python creates one anyway in the dict, but hashlib.md5 should work:

    >>> from hashlib import md5
    >>> checksum=md5('[5206081:5207221](-)'.encode('utf-8'))
    >>> checksum.hexdigest()
    '4dbae8825761207cab29f6dcdf97785a'
    >>> checksum=md5('[1964876:1965035](-)'.encode('utf-8'))
    >>> checksum.hexdigest()
    'a6d96d4f1f461226fefeb93ea67e110b'

### DIFF
--- a/antismash/common/secmet/record.py
+++ b/antismash/common/secmet/record.py
@@ -15,7 +15,7 @@ import bisect
 from collections import Counter, defaultdict, OrderedDict
 import logging
 from typing import Any, Dict, List, Optional, Sequence, Type, Tuple, Union, cast
-from zlib import crc32
+from hashlib import md5
 
 from Bio import SeqIO
 from Bio.Seq import Seq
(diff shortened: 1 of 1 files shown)


kblin:
Is this also failing in current master? We switched the logic after running into some conflicts ourselves. The reason we don't want to use md5 or something else generating longer hashes is that we need to use these IDs for additional CDS features we create.

kblin:
Oh, and of course using only the first X characters of a longer hash would run into the same issues, at higher computational costs.

SJShaw:
#422 reduced the number of comparisons for collisions in a genome from `n(n-1)` (`n` being the number of CDS features) to `m-1` (where `m` is the number of CDS features with the same name). This is in the order of millions of times less likely to cause a collision.

This is a tiny script that sets up the scenario above. With the merge of #422 there's no issue, but the commit before that  does have the collision. Changing them to have the same locus tag to force the collision results in both master and this PR correctly rejecting these as having duplicate CDS names.

```
from antismash.common.secmet import Record, CDSFeature, FeatureLocation

first_loc = FeatureLocation(1964876, 1965035, 1)
first = CDSFeature(first_loc, locus_tag="A", translation="A" * (len(first_loc)//3))

second_loc = FeatureLocation(5206081, 5207221, 1)
second = CDSFeature(second_loc, locus_tag="B", translation="B" * (len(second_loc)//3))

record = Record()
record.add_cds_feature(first)
record.add_cds_feature(second)
```

This PR will make collisions less likely still, but I think the issues with new name length (mentioned earlier) negate that improvement.

jlv100:
> Is this also failing in current master? We switched the logic after running into some conflicts ourselves. The reason we don't want to use md5 or something else generating longer hashes is that we need to use these IDs for additional CDS features we create.

No, this passes in latest dev docker image.

jlv100:
Looks like this case is solved in #422 so I will close this PR.  Thanks.

PR #436, opened 2022-04-05
Minor fixes

SJShaw:
Changes some leftover wording of 'cluster' to 'region' in outputs.

Simplifies FASTA generation for prodigal input (incidentally fixes a crash that could occur with seqio including spaces in outputs for some very specific inputs).


kblin (approved):
Looks good to me 👍

PR #444, opened 2022-05-09
Fix for missing cross-CDS NRPS/PKS modules in bubble visualisation

SJShaw:
The classification system for NRPS/PKS genes carried over from antiSMASH 4 did not take into account the existence of cross-CDS modules and was too broad with the glycopeptide classifications.

The missing PKS classification for some genes, combined with an ordering issue in the chain detection within the order finding code led to some genes with split CDS modules being skipped in the domain docking analysis and the resulting bubble visualisation.

This PR completely removes the old, incorrect classification system. Order finding now uses the presence of complete modules instead of the classification system. 
It also includes a couple of other minor fixes cleaning up unused variables, arguments, and imports, along with fixing `make combined-coverage` only using the integration coverage (this did work with older versions, but it stopped working at some point and wasn't noticed).


kblin (approved):
Looks good to me 👍

PR #445, opened 2022-05-09
RiPP precursor detection fixes

SJShaw:
The `all_orfs` module was doing gene finding while trying to find the longest possible gene. After those genes were found, they would then be culled if any overlapped with existing genes. This seems to have culled the vast majority of possible thiopeptide precursors (along with other RiPP) modules.

This PR changes the `all_orfs` module to limit the search space to only existing intergenic areas (with customisable padding and size limits), increasing the number of hits and reducing typical size. However, this size reduction isn't accurate enough for some thiopeptide precusors, so the existing cleavage site predictor was used within the new genes from `all_orfs` to trim to the nearest start codon. This trimming function is now part of `all_orfs`, and at some stage the other RiPP modules can do something similar.

All RiPP modules using `all_orfs` have been changed to use the newly available size limits and reduced the magic number issue with the size constraints. Only the thiopeptides module is using the new trim functionality.


kblin (approved):
Looks good to me 👍

PR #459, opened 2022-07-01
common.subprocessing: make muscle version check case insenstive

### LINKED ISSUE #457: MUSCLE prereq checking doesn't match most recent version
**Describe the bug**
When you run `antismash --check-prereqs` one thing is does it run this function

```
def run_muscle_version() -> str:
    """ Get the version of the muscle binary """
    muscle = get_config().executables.muscle
    command = [
        muscle,
        "-version",
    ]

    version_string = execute(command).stdout
    if not version_string.startswith("MUSCLE"):
        msg = "unexpected output from muscle: %s, check path"
        raise RuntimeError(msg % muscle)
    # get rid of the non-version stuff in the output
    return version_string.split()[1]
```
However, the current version of MUSCLE does not have this expected output. The output is lowercase and not uppercase. 
```
$ muscle -version
muscle 5.1.linux64 []
Built Feb 24 2022 03:16:15
```
What version of MUSCLE should I use? Alternatively, could the way MUSCLE is checked be changed?

SJShaw:
Fixes #457

### DIFF
--- a/antismash/common/subprocessing/muscle.py
+++ b/antismash/common/subprocessing/muscle.py
@@ -50,7 +50,7 @@ def run_muscle_version() -> str:
     ]
 
     version_string = execute(command).stdout
-    if not version_string.startswith("MUSCLE"):
+    if not version_string.lower().startswith("muscle"):
         msg = "unexpected output from muscle: %s, check path"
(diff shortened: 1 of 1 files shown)


kblin (approved):
Looks good to me 👍

PR #460, opened 2022-07-05
Misc fixes

kblin:
* HMM detail extraction script
* Fix for an assert triggered by `run_mibig.py` so the log contains a useful error message.


kblin:
duh, forgot to rebase

PR #466, opened 2022-07-27
MUSCLE5 error

Marios-Nikolaidis:
Added a runtime error when the default muscle instalation is muscle5, since it does not have the -profile option


SJShaw:
MUSCLE has been removed as a dependency as of #508, so I'll close this PR

PR #467, opened 2022-08-09
secmet: handle bad NCBI misc features for Pfam hits

SJShaw:
Some NCBI annotations of Pfam hits can end up with very strange locations, e.g. this from `AM743169.1`, where the parent CDS location is `complement(join(3683384..3684415,3684417..3686921))`
```
     misc_feature    complement(order(3683871..3684154,3683813..3684415))
                     /locus_tag="Smlt3626"
                     /inference="protein motif:HMMPfam:PF02518"
                     /note="HMMPfam hit to PF02518, Histidine kinase-, DNA
                     gyrase B-, and HS, score 3.6e-44"
                     /pseudo
```

This PR removes any of these redundant exons to avoid triggering the origin-crossing detection logic.

### DIFF
--- a/antismash/common/secmet/locations.py
+++ b/antismash/common/secmet/locations.py
@@ -467,3 +467,31 @@ def frameshift_location_by_qualifier(location: Location, raw_start: Union[str, i
         codon_start *= -1
 
     return _adjust_location_by_offset(location, codon_start)
+
+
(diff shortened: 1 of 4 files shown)


kblin (approved):
Looks good to me 👍

PR #468, opened 2022-08-09
cluster_compare: don't create HTML content for search types with no results

SJShaw:
This reliably happens when a region contains no protoclusters, but the comparison config includes one of the protocluster search modes (e.g. via sideloading). The javascript data already has this condition, which led to javascript errors attempting to access missing information because the selector still existed.

Existing code in the HTML template handles the case for when all search modes have no results.


kblin (approved):
Looks good to me 👍

PR #471, opened 2022-09-05
Misc fixes

SJShaw:
Fixes a few minor issues:
- introduces a cap of 100,000 aminos for translations, which is the maximum for external tools like hmmscan
- prevents genefinding from calling a gene with more than 90% of the new cap (since it was including half megabase stretches of `N` in some odd assemblies)
- allows `--sideload-simple` to specify an end-point past the record end, automatically truncating it to the record end if necessary
- corrects cross-CDS module parent name ordering to match the order of domains provided, rather than alphabetically sorted


kblin (approved):
Looks good to me 👍

PR #473, opened 2022-09-21
Misc fixes

SJShaw:
Adds functionality to discard corrupted antiSMASH-specific feature types to handle parsing failures (e.g. an `aSModule` with no relevant qualifiers). Missing qualifiers raising `KeyError` (deliberately or not) will not be caught by this system.

Fixes a number of issues:
- ClusterBlast SVGs would have the last hit missing colours due to skipping the same accession as the input _after_ the colours were generated
- the new intergenic area search for finding all ORFs was using the wrong directionality for padding, and related tests weren't correctly checking values
- improves testing of split cross-origin features, previously only the original biopython feature was being tested


kblin (approved):
Looks good to me 👍

PR #478, opened 2022-10-16
hmm_detection: Add detection of NRP metallophores

zreitz:
The existing antiSMASH `siderophore` cluster type only recognizes NRPS-independent siderophore synthetase (NIS) genes with the `IucA_IucC` pHMM.  The `NRP-metallophore` rule finds the other major type of metallophore, where NRPS genes are colocalized with metallophore-specific biosynthesis genes (~20% of all NRPSs).

Notes:
* `NRP-metallophore` is in the `NRPS` category, and a `not` rule was added to `NRPS` to avoid cross matching. 
* `siderophore` and `NRP-metallophore` now use the `metallophore` CSS mix-in (same red color). More metallophore rules are planned.
* Compared with 758 manually annotated NRPS clusters, the rule had 94% accuracy, 97% precision, and 78% recall. Known false negatives: these rules cannot detect siderophore clusters where the precursor/accessory genes are missing or elsewhere in the genome.
* A manuscript describing the rules and results on representative genomes is in preparation.
* Alignments used to make the pHMMs are available here: https://github.com/zreitz/antismash/tree/sid-detection-testing/NRP-metallophore-HMM-alignments

**The rule looks for the following pHMMs:**

* `VibH_like` and `Cy_tandem` are specialized condensation domains (10.1016/j.bbapap.2017.05.010) found in siderophore BGCs. Adenylation domains are not (always) on the same CDS, and these are the only exceptions to the requirement for `cds(Condensation and AMP-binding)`. 
  
* `IBH_Asp`, `IBH_His`, `TBH_Asp`, `CyanoBH_Asp2`, and `CyanoBH_Asp2` are Asp/His beta-hydroxylases (10.1073/pnas.1903161116 and 10.1128/AEM.03128-20). `SBH_Asp` is a negative condition; this subclade of `IBH_Asp` is found only in syringomycin-like phytotoxins.
  
* `IPL` (isochorismate pyruvate lyase, 10.1074/jbc.M605470200) and `SalSyn` (salicylate synthetase, 10.1128/JB.187.15.5061-5066.2005) are two methods of salicylic acid biosynthesis.
  
* `EntC` and `EntA` are the first and third/final step of 2,3-dihydroxybenzoate biosynthesis (10.1039/c7mt00111h). An EntB HMM wasn't required and adding it would cause more false negatives.

* `GrbD` and `GrbE` are required for graminine biosynthesis (10.1002/anie.201906326).
  
* `FbnL` and `FbnM` are required for DMAQ biosynthesis (10.1021/acschembio.0c00809).
  
* `PvdO` and `PvdP` are required for pyoverdine chromophore formation (10.1128/JB.01376-13).

* `Orn_monoox` and `Lys_monoox` hydroxylate the sidechain N of Orn and Lys, respectively. A negative condition for `KtzT` filters NRPs where N5-OHOrn is converted to piperazate instead of a chelating group (10.1038/nchembio.2411), and `MetRS-like` filters N-N bonds (10.1038/nchembio.2411).
  
* `VbsL` is a racemase involved in vicibactin biosynthesis (10.1021/ja9056008). Orn N-hydroxylase VbsA was not chosen because it is too similar to IucD and may cause false positives.


SJShaw:
The `not` in the `NRPS` rule isn't going to stop it from appearing, especially since the cutoff for the new rule is larger (though it shouldn't be, looking at BGC413, the largest gap between core genes is less than 10kb, rather than 30kb as per the comment in the rule). Even then, if a gene with a complete NRPS module is at the opposite end of a protocluster from hits of these new siderophore profiles, it'll just move the NRPS to the side. Then, when the `NRPS` walks back, it'll allow the `NRPS-like` rule to match, which means BGC413 starts looking like this:
image

Using the `SUPERIORS` section of the rule to prevent these protoclusters from overlapping would be a better solution. Unfortunately, that doesn't support chaining at the moment because we were worried that (in general) they weren't transitive enough to justify it, but we might have to revisit that.

SJShaw:
> Yikes, I clearly did not test that last commit enough. `SUPERIORS` does sound ideal. My motivation for avoiding cross-detection is to make these clusters red in the cluster overview to match siderophores; I don't mind it in the individual cluster tabs. Are there any currently implemented options for that? I couldn't figure out how to override NRPS/NRP-metallophore hybrids being green.

Any hybrid formed from rules with different `CATEGORY` values are that same light-blue colour. Hybrids formed from rules with the same `CATEGORY` will use that category's colour (in `antismash/outputs/html/secmet.scss`). That category colour is also what's used if there's no specific CSS class defined for the protocluster/rule name.

> Setting cutoff to 20kb drops the right-most NRPS in BGC413. I assume because it's no longer within the cutoff from `PvdO` (closest small core gene), but I don't understand why the other NRPS genes don't count: image

Ah, of course, I didn't think of that. It's fine to have 30KB then. In the worst case it'll form a neighbouring candidate cluster and users can just merge them.

> ...
> 
> I suppose my expectation for `a and b and not c` was more `a and b and discard the whole protocluster if you find c`

That is effectively how it works, it's just that the negative requirements (`not c`) need to be found within the cutoff distance of the CDS(es) fulfilling the positive requirements (`a and b`). Otherwise it would need a second, larger cutoff distance for the negative conditions, and that gets messy enough that we added the `SUPERIORS` setup to handle it instead.

SJShaw:
While I think of it, the `SUPERIORS` would need to be _in addition_ to the `not` in the NRPS rule, otherwise that case with the trailing NRPS protocluster would end up with the NRPS being lost.

SJShaw:
I've resolved the merge conflict, added `EXAMPLE` lines for the two referenced compounds, and squashed down the commits. It should otherwise be identical to what you had, but if you want to double check, feel free.

Otherwise, I think this is good to go. If at some point we want to add the `SUPERIORS` tag for `NRPS` protoclusters, the transitivity issue is now resolved, but I think it best to keep it as is for now until we get a good feel for how often the more awkward overlaps happen.

SJShaw (approved):

zreitz:
Looks good to me! I still want to add a few other families of metallophores closer to the deadline, but I need to organize a few things. I'll make a new pull request when it's ready. They'll be easy, `A + B` types.

PR #480, opened 2022-10-27
nrps_pks: Add PARAS, untested.

BTheDragonMaster:
Introduces PARAS A-domain predictions. Dos not yet include tests. Does not use PARAS predictions yet as consensus.

### DIFF
--- a/antismash/modules/nrps_pks/paras.py
+++ b/antismash/modules/nrps_pks/paras.py
@@ -1,0 +1,65 @@
+from typing import List, Tuple, Dict, Any
+
+from .data_structures import Prediction
+from antismash.common.html_renderer import Markup
+from antismash.detection.nrps_pks_domains import ModularDomain
+from paras.run import run_paras as paras
+
+
+class ParasResult(Prediction):
+    """ Holds all the relevant results from PARAS for an adenylation domain """
+    def __init__(self, paras_predictions: List[Tuple[float, str]]) -> None:
+        super().__init__("PARAS")
+        self.predictions = paras_predictions
+        self.best_prediction = self.get_best_prediction()
+
+    def get_classification(self) -> List[str]:
+        if self.predictions:
+            predictions = [self.predictions[0]]
+            for prediction in self.predictions[1:]:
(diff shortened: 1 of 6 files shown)


SJShaw:
I know this is a draft, so I won't comment on but it seems like the non-antiSMASH portions of PARAS need some fixes.
This warning is repeated twice for every input, I remember discussing ways of handling it with you back in September.
```
lib/python3.9/site-packages/sklearn/base.py:329: UserWarning: Trying to unpickle estimator DecisionTreeClassifier from version 0.24.2 when using version 1.1.3. This might lead to breaking code or invalid results. Use at your own risk. For more info please refer to:
https://scikit-learn.org/stable/model_persistence.html#security-maintainability-limitations
```
Since the unpickle is happening for every domain, that's a lot of repetitive loads from disk. Does PARAS support a batch mode where it only has to load once for any given set of domains?

There's also some debug printing coming from PARAS that we don't want. The records I ran on crashed before reaching RAIChU, so I can't comment on that part.

Neither PARAS or RAIChU have a license, that's also something that we'd like to know about.

SJShaw:
> Thanks for the feedback. Could you tell me 1. which records you ran RAIChU on and 2. which error messages you got, so that I can debug on my end (also useful for future PRs)?

The first was balhymicin (Y16952), which included as test data in antiSMASH. It's the same issue as every other input I tried, with variations of prediction substituted:
```
$ ./run_antismash.py antismash/test/integration/data/Y16952.gbk
 ...
 File "antismash/modules/nrps_pks/parsers.py", line 74, in generate_nrps_consensus
  raise ValueError("%s generated bad prediction string: %s" % (method, best))
ValueError: PARAS generated bad prediction string: "3,5-dihydroxyphenylglycine"
```

I assumed you were still working on that part, which is why I didn't go into detail about it.

kblin:
This will be handled via the upcoming experimentalSMASH infrastructure for the time being, close for now.

kblin:
Testing to see if I can get this to work with experimentalSMASH

kblin:
Superseded by the PARAS outlink added in #761

PR #482, opened 2022-10-28
Misc fixes

SJShaw:
A number of small fixes and updates:
- makes path tests resolve symlinks to avoid tests failing on Mac when using `TemporaryDirectory`
- removes the absolute path from `make unit` for Mac users with old(/as recent as apple provides) BASH versions
- removes two cases where created Protocluster could be on the negative strand (this has no real impact, but good for consistency)
- replaces the now-defunct Pfam URLs with the new host at EBI
- removes the restriction on detection rules specifying a superior that itself also had a superior
- adds a tiny bit of padding for details panel tabs in HTML to improve clearance for tall characters near the start or end


kblin (approved):
Looks good to me 👍

PR #487, opened 2022-11-01
Detection: Add improved phosphonate detection

malanjary-wur:
Improved Phosphonate detection model, added secondary core gene models, and tighter cluster neighbourhood (work of Artem Fetter). Previous model moved to loose detection rule.


antismash/detection/hmm_detection/data/hmmdetails.txt
```diff
@@ -151,6 +151,25 @@ PF07366	SnoaL family	20	PF07366.hmm
...
+phosphonatesMeIsocitrateLyase	Methylisocitrate Lyase	250	phosphonatesMeIsocitrateLyase.hmm
+2-Hacid_dh_C	D-isomer specific 2-hydroxyacid dehydrogenase, NAD binding domain	25	2-Hacid_dh_C.hmm
+Aldedh	Aldehyde dehydrogenase family	23	Aldedh.hmm
+Aminotran_1_2	Aminotransferase class I and II	19	Aminotran_1_2.hmm
+Aminotran_5	Aminotransferase class-V	21	Aminotran_5.hmm
+AurF	P-aminobenzoate N-oxygenase AurF	29	AurF.hmm
+CTP_transf_like	Cytidylyltransferase-like	21	CTP_transf_like.hmm
+DUF4992	Domain of unknown function	29	DUF4992.hmm
+Fe-ADH	Iron-containing alcohol dehydrogenase	26	Fe-ADH.hmm
+HMGL-like	HMGL-like	29	HMGL-like.hmm
```
SJShaw, line 165: This is a duplicate name of `HGML-like` earlier in the file, which is the same Pfam (though a slightly older version). It would be better to update that file than have two copies, but we'll need to check it doesn't impact any other rule using it (currently only `betalactone`).

antismash/detection/hmm_detection/cluster_rules/strict.txt
```diff
@@ -503,8 +503,8 @@ RULE phosphonate
     CATEGORY other
     DESCRIPTION Phosphonate
     CUTOFF 20
-    NEIGHBOURHOOD 20
-    CONDITIONS phosphonates
+    NEIGHBOURHOOD 5
+    CONDITIONS (phosphonatesV2 and (2-Hacid_dh_C or Aldedh or Aminotran_1_2 or Aminotran_5 or AurF or CTP_transf_like or DUF4992 or Fe-ADH or HMGL-like or Metallophos_2 or NTP_transf_3 or NTP_transf_5 or PALP or PEP_mutase or TPP_enzyme_C or TPP_enzyme_N or phosphonatesV2) and not phosphonatesHydrolase and not phosphonatesMeIsocitrateLyase)
```
SJShaw, line 507: The second section of this condition will always be satisfied if the first section is, because it's including the first condition as an `or` (i.e. `A and (A or ...)` makes all of the other included 'or' terms irrelevant. Is this intentional?

Also, it would probably be better to move the old `phosphonates` profile to a new name and have just `phosphonates` instead of the `V2` version.

antismash/detection/hmm_detection/cluster_rules/loose.txt
```diff
@@ -29,3 +29,10 @@ RULE halogenated
     CUTOFF 5
     NEIGHBOURHOOD 10
     CONDITIONS Trp_halogenase
+
+RULE phosphonate-loose
```
SJShaw, line 33: Without the `SUPERIORS` field set to the strict `phosphonate`, this rule will likely duplicate all or most of the stricter rule.

SJShaw (commented):
Along with the other small issues, it might be nice to have an example cluster or two to help highlight the difference between the rules.

antismash/detection/hmm_detection/data/hmmdetails.txt
malanjary-wur: Resolved by removing the duplicate HMGL-like as it is still detected using the previous model

antismash/detection/hmm_detection/cluster_rules/strict.txt
malanjary-wur: Renamed the new rule. Repeating the phosphonates model was intentional, the additional models are not required but allow tighter neighbourhood value for most clusters (i.e. a cluster with only the phosphonate hit will still be detected)

antismash/detection/hmm_detection/cluster_rules/loose.txt
malanjary-wur: Thanks for this - SUPERIORS has been set here. This also was a solution for "other" (catch all) rule that was calling hybrids for some phosphonate clusters

kblin: I think that's the wrong fix for the `other` rule. As far as I can see what's causing the hybrid calls is the `fom1` profile, and that should just move to the phosphonates cluster rule instead. That profile is even called after one of the genes in your example cluster. :)

malanjary-wur: Ok agreed thanks, I will just remove the fos1 model from the 'other' rule and remove the SUPERIORS flag. This is currently being hit with the updated phosphonate model so should not be missed.

kblin: yeah, I guess that'd also be fine.

malanjary-wur: Great, I just now pushed the changes and removed the fom1 model as it was not being used by any other rule

SJShaw (approved):

PR #489, opened 2022-11-03
Misc fixes

SJShaw:
Fixes the thiopeptides module schema results having been missed in the CompaRiPPson update (only affected runs with a version post-b8eff304 (merged a day or so ago), with results from prior to b8eff304, merged a day or so ago. Future runs with old results will work as expected.

Fixes `antismash.common.all_orfs.find_all_orfs()` having incorrect behaviour when provided a specific area to search in. This only affected RiPP modules when searching for unannotated precursors. The precursors found would have existed, but their locations within the area were incorrect.

Updates pylint exclusions to remove obsolete message types and add new exclusions for message types which are commonly false positives (e.g. the test files doing equality checks against a specific empty type is important and shouldn't be simplified down to a simple "falsey" value).

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,7 +1,7 @@
 [MASTER]
 ignore=external
 reports=y
-disable=bad-continuation,too-few-public-methods
+disable=too-few-public-methods
 
 [MESSAGES CONTROL]
 # raising 'from' is almost entirely false positives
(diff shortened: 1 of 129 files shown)


kblin (approved):
Looks good to me 👍

PR #492, opened 2022-11-10
Fix some database dir HMMs not being repressed when necessary

SJShaw:
Fixes an edge case where one was doing development work with `make clean` and `--database-dir` was using the `databases` directory in the source tree rather than in another location.


kblin (approved):
Looks good to me 👍

PR #493, opened 2022-11-11
Add interactive gene overview table to HTML output

SJShaw:
### Gene table

Each row in the table contains most of the information available in the existing focus panel (top right of the region view in HTML), normally shown when clicking an individual gene in the drawing.

Main features:
- filterable/searchable rows, based on visible row columns along with additional hidden data matching the query term (using regex in the query is also an option)
- automatic zooming of region drawing to the filtered/selected genes (a toggle switch is included to disable this)
- automatic scrolling of the table to show selected genes (starting at the first, in cases where many genes are selected)
- genes selected in the drawing will also include a marker on the left of the row to help identify the relevant row(s)

Adding more data to the filter targets is relatively easy, but will need a paired update for javascript and visualiser module. Data available for searching currently:
- CDS name (as used in the rest of antiSMASH output when a single identifier is used)
- product annotation (not supplied by antiSMASH)
- primary gene function (as per the region drawing)
- secondary gene function annotations (annotated by antiSMASH, including SMCoG hits, resistance status, and so on)
- names of biosynthetic profiles found within the domain which were used in cluster detection

An explicit export option isn't provided, but browsers can easily select desired rows and copy/paste directly into typical spreadsheet software.

A demo output is available [here](https://antismash.secondarymetabolites.org/upload/prototypes/gene_table/index.html).

### Other changes
- added a wildcard system to allow javascript to fill in data within the HTML, reducing data deduplication (the demo above would only be an increase of about 30kb in combined HTML/JSON size with this change)
- HTML generation now strips leading whitespace and blank lines from the generated page (effectively reducing the net size increase of 30kb for the demo above to zero)
- fixes the `#defs` SVG used in the HTML having a non-zero size, despite not being visible, removing some empty scroll space
- adds a initial pass at a reusable construct for indicating which genes are selected in the region drawing
- some other small renderer helpers and fixes


kblin (approved):
Looks good to me 👍

PR #494, opened 2022-11-14
Database updates

SJShaw:
- Fixes an issue brought in by #485 that commented out most databases.
- Adds better handling to prerequisite checks for building docker containers where the database directory is mounted at runtime.
- Adds some tests to ensure that trying to prepare data in a runtime mounted directory is caught early on.


kblin (approved):
Looks good to me 👍

PR #505, opened 2022-11-29
Update NAPDOS links to v2, and other small changes

SJShaw:
Also includes a couple of small fixes for test file creation and deduplication of domain sequences in NRPS/PKS domain tooltips.


kblin (approved):
Looks good to me 👍

PR #506, opened 2022-11-30
Add A domain predictions from PARAS

kblin:
Cleaned up version of #480


kblin:
This will be handled via the upcoming `experimentalSMASH` infrastructure for the time being, close for now.

PR #508, opened 2022-12-05
Remove MUSCLE as a dependency, replacing it with brawn or hmmpfam2

SJShaw:
MUSCLEv5 removes the profile-profile mode (mentioned in a partial fix in #466), which was used in the `nrps_pks` and `smcog_trees` modules. For the main NRPS/PKS use case of extracting signatures for A-domains, using hmmpfam2 as a replacement was an accuracy improvement. For all other cases brawn was used as a replacement alignment tool.

Initial install/data-preparation time increases from around 6 seconds to around 20 seconds, since brawn is building its cache files. Once caches are built, the time to check prerequisites increases from 1.0s to 1.3s.

Runtime is largely the same. Depending on the input, using brawn ends up slightly faster or slightly slower.


kblin (approved):
Looks good to me 👍

PR #512, opened 2022-12-06
nrps_pks: remove a stray print from stachelhaus

SJShaw:
I thought I'd caught this in the previous PR and removed it, but I'd removed a _different_ print.


kblin (approved):
Looks good to me 👍

PR #513, opened 2022-12-06
NRPS/PKS preparation work for transATor inclusion

SJShaw:
Changes NRPS/PKS qualifier and modular domains to allow for arbitrary depths of sub-types.


kblin (approved):
Looks good to me 👍

PR #474, opened 2022-09-23
Added trans at hmms

FriederikeBiermann:
Transatksdomains-> hmms for all clades of ks domains additionaldomainstransatpks -> additional domains that are not present in the nrpspksdomains and that are not correctly detected by those hmms Testing done: 
ran all hmms (original ones+ new ones) on a test set of 50 trans-at-pks clusters + 50 cis-at-pks-clusters -> correct hmms still score highest scores but: trans-at-pks-out hmm seems to score sometimes quite similar to cis-at-ks -> in all cases still lower, but similar locations of detected domains seem to be quite similar (sometimes 1-5 aa difference)

PR #517, opened 2022-12-07
TransATor profile detection

SJShaw:
A cleaner, less involved version of #474


kblin (approved):
Looks good to me 👍

kblin (approved):
Looks good to me 👍

PR #521, opened 2022-12-08
Small fixes

SJShaw:
- fixes cross-CDS modules on reverse strand CDSes from being flipped in the visualisation
- fixes "Selected features only" for NRPS/PKS domains in the HTML output not respecting existing selections on first load
- fixes CompaRiPPson data not being checked by all RiPP modules
- silences logging of CompaRiPPson data preparation/checking on checks after the first


kblin (approved):
Looks good to me 👍

kblin (approved):
Looks good to me 👍

PR #523, opened 2022-12-09
nrps_pks_domains: fix default state of modules showing domains

SJShaw:
In 4ece39d74 I changed the name of the NRPS/PKS substrate consensus key, but didn't update it in the drawing logic, which disabled, by default, the showing of module lids with substrates even when there were substrates predicted for modules.

This PR fixes that and changes the tests to catch similar issues in future.


kblin (approved):
Looks good to me 👍

PR #524, opened 2022-12-13
html: remove antiSMASH javascript from the source tree

SJShaw:
As the size of the shipped javascript increases and the frequency of updates increases, including it in the repository is getting to be more and more silly. This PR moves the javascript out of the source tree completely.

If no other actions are taken on an install, the generated HTML results will point to javascript via the normal antiSMASH download host.
This can be overridden by running the `download-antismash-databases` command/script, which will keep a copy in the databases directory and that copy will be included in all generated results just as is done currently.
The database override can itself be overridden by placing a copy of `antismash.js` in the tree at `antismash/outputs/html/js/antismash.js` (this has been added to `.gitignore`).

This also allows for results referencing the remote URL to get any javascript patches without having to touch this repository, though this only applies to patch version changes. Major and minor version changes will not be automatic, as those are likely to break compatibility.
However, installs using containers or otherwise frozen versions using the database download won't get those patches, as they'll download the latest at time of install/building.


kblin (approved):
Looks good to me 👍

PR #525, opened 2022-12-15
Small HTML tweaks and a version bump to beta

SJShaw:
Updates HTML tooltips for:
- NRPS/PKS tabs to include links to the module visualisation(s) docs page (and point at an existing rule for PKSs)
- TFBS to link to the relevant docs page
- RiPP modules to link to the CompaRiPPson docs page

Also tweaks CompaRiPPson HTML output to sort groups by descending similarity, then descending group size, then accession name. This makes more common sequences more obvious.

Also bumps the version to "beta" state.

### DIFF
--- a/antismash/common/comparippson/__init__.py
+++ b/antismash/common/comparippson/__init__.py
@@ -6,6 +6,7 @@ import os
 from typing import List
 
 from antismash.common import path, subprocessing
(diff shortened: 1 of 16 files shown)


kblin (approved):
Looks good to me 👍

PR #530, opened 2023-01-19
PKS order finding tweaks

SJShaw:
Fixes finding a final gene for the order when thioesterases and other terminal domains were in a cross-CDS module.

Tweaks the order finding itself to bias towards keeping consecutive genes together.

### DIFF
--- a/antismash/modules/nrps_pks/orderfinder.py
+++ b/antismash/modules/nrps_pks/orderfinder.py
@@ -197,8 +197,9 @@ def find_first_and_last_cds(cds_features: List[CDSFeature]) -> Tuple[Optional[CD
     # find the end
     for cds in cds_features:
         if cds.modules and cds.modules[-1].is_final_module():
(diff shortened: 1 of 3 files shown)


antismash/modules/nrps_pks/orderfinder.py
```diff
@@ -362,6 +393,9 @@ def rank_biosynthetic_orders(n_terminal_residues: Dict[str, str],
         score = 0
         interactions = [order[i:i + 2] for i in range(len(order) - 1)]
         for gene, next_gene in interactions:
+            print(gene.get_name(), next_gene.get_name(), "but", following_genes.get(gene.get_name()))
```
kblin, line 396: Debug output?

kblin (changes requested):
I think a debug `print()` snuck in here, fine otherwise.

kblin (approved):
Looks good to me 👍

PR #535, opened 2023-01-27
hmm_detection: Add dectection rule for fungal arginine-containing cyclic dipeptides

kblin:
This is based on doi: 10.1038/s41589-022-01246-6

Marking as draft until the last three AnkA homologs are unembargoed so I can use all the sequences for the RCDPS profile.


kblin:
now uses all the known RCDPS  sequences for the profile.

SJShaw (approved):

PR #538, opened 2023-02-02
Fixes for rule superior/inferior edge cases

SJShaw:
A example was discovered of genes being marked as core genes without being contained by a protocluster core, like so:
example
The gene above overlaps with it's neighbour by 10 bases. It's considered a core gene for a protocluster type not present, but both of the protoclusters present are superiors to the missing protocluster.

There were two separate issues causing this to happen, both of which are handled by this pull request.

1. It was possible for a protocluster to be lost during candidate cluster creation, due to the deduplication process using only coordinates. This meant that the interleaved candidate that would have contained the missing protocluster was being dropped, orphaning the protocluster and resulting in it being discarded.
This was fixed by including the product list in the uniqueness check. On top of the tests, an assertion now ensures that protoclusters can't be lost in this way again without being noticed.

2. It was possible for a protocluster with an adjacent superior to be discarded when the adjacent genes overlapped slightly, as in this case with a 10 base overlap. This was fixed by moving from a checking if the locations overlapped to checking that the core CDS features overlapped.

### DIFF
--- a/antismash/common/secmet/features/candidate_cluster/formation.py
+++ b/antismash/common/secmet/features/candidate_cluster/formation.py
@@ -31,8 +31,8 @@ def create_candidates_from_protoclusters(protoclusters: List[Protocluster]) -> L
     if not protoclusters:
         return []
 
-    # ensure that only one candidate cluster can be created for each specific start/end combination
-    existing: Set[Tuple[int, int]] = set()
+    # ensure that only one candidate cluster can be created for each specific combination
+    existing: set[tuple[int, int, tuple[str, ...]]] = set()
 
     def build_candidates(groups: List[List[Protocluster]], kind: CandidateClusterKind) -> List[CandidateCluster]:
         """ Creates candidates of the given kind, one for each group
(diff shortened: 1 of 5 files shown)


antismash/common/secmet/features/candidate_cluster/formation.py
```diff
@@ -32,7 +32,7 @@ def create_candidates_from_protoclusters(protoclusters: List[Protocluster]) -> L
         return []
 
     # ensure that only one candidate cluster can be created for each specific start/end combination
-    existing: Set[Tuple[int, int]] = set()
+    existing: set[tuple[int, int, tuple[str, ...]]] = set()
```
kblin, line 35: The comment is slightly out of sync now

kblin (changes requested):
LGTM, apart from that one comment

antismash/common/secmet/features/candidate_cluster/formation.py
SJShaw: Good catch, fixed.

kblin (approved):
Looks good to me 👍

PR #540, opened 2023-02-03
Misc fixes

SJShaw:
- fixes up some typos and missing compounds in detection rules
- fixes bad arg order leading to seeds being used as a start coordinate
- trims off fractions from molecular weight estimates for Type II PKS results
- fixes TransAT profiles not properly respecting `logging_only` mode, which crashed `--check-prereqs`
- changes CSS for cluster compare's table headers so that chromium-based browsers display the same as firefox


kblin (approved):
Looks good to me 👍

PR #544, opened 2023-02-10
chore: handle more workflow deprecations

SJShaw:
Also explicitly checks out the merge commit for linting PRs.


kblin (approved):
Looks good to me 👍

PR #546, opened 2023-02-13
config: add a '--prepare-data' alias for '--check-prereqs' option

SJShaw:
For being more explicit about it's possible use in documentation, an alternative name would be useful. The old name needs to stay for legacy reasons as well as being a valid name for checking prerequisite binaries.


kblin (approved):
Looks good to me 👍

PR #548, opened 2023-02-20
Misc fixes

SJShaw:
- Fixes database download script not respecting `executable_paths` option, if defined, in antiSMASH config files
- Fixes the `smcog_trees` module not respecting `fasttree` when defined in `executable_paths`
- Changes "PFAM" to "Pfam" in CDS focus panel, since that's the correct casing when not using an accession
- Shows monomer predictions tab for NRPS/PKS regions if any predictions were made, previously this would be hidden along with polymer predictions if no polymers were present (i.e. when no complete NRPS/PKS modules were detected in the region)


kblin (approved):
Looks good to me 👍

PR #572, opened 2023-03-21
nrps_pks: Go with best 8 Å match if multiple Stachelhaus matches for the same stach signature exist

kblin:
This fixes the many unclear predictions for e.g. the CDA cluster.


kblin:
Ok, the convention issue is that one of my test names is too long, I don't see a refactor issue count change locally, so I have no idea what that'd be.

kblin:
looks like the refactor issue just vanished again :shrug:

SJShaw (approved):

PR #576, opened 2023-03-24
Further fixes for lost protoclusters

SJShaw:
In some rare circumstances, it was possible for protoclusters to be lost. This typically happened on contig edges, when a cluster was considered "interleaved" but only had a very minor overlap, such that a candidate that had the same coordinates as another would've been created but was being dropped, which also potentially dropped one or more protoclusters.

This PR fixes the known cases of these lost protoclusters by "promoting" any orphaned protocluster into the more strict type that already existed at those coordinates. For these protoclusters that are promoted, a SINGLE candidate cluster will also be created, unless there's already another stronger candidate cluster at those specific coordinates as well (likely only to happen on contig boundaries).

A related change is that it is no longer true that all protoclusters in a NEIGHBOURING candidate will have a matching SINGLE. In the case that the SINGLE's coordinates already exist as a candidate (e.g. for a neighbouring case where a small protocluster is entirely contained within the neighbourhood of a larger protocluster), it will not be created.

Also fixes/improves a couple of repr strings for related classes.

### DIFF
--- a/antismash/common/secmet/features/candidate_cluster/formation.py
+++ b/antismash/common/secmet/features/candidate_cluster/formation.py
@@ -32,20 +32,38 @@ def create_candidates_from_protoclusters(protoclusters: List[Protocluster]) -> L
         return []
 
     # ensure that only one candidate cluster can be created for each specific combination
-    existing: set[tuple[int, int, tuple[str, ...]]] = set()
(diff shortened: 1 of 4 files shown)


kblin (approved):
Looks good to me 👍

PR #580, opened 2023-03-28
add TFBS_finder hits to gbk file

HAugustijn:
Adds the tfbs_finder results to the GenBank file. Example record:

misc_feature    621641..621659
               /note="TFBS match to DmdR1, Iron(II)-dependent repressor,
                confidence: strong, score: 36.53"


kblin:
There seems to be an unrelated patch that changes the GC percentage calculation data type from `str` to `Seq` mixed in, could you make this a separate patch?

PR #585, opened 2023-04-18
TFBS fixes/updates

SJShaw:
Includes what was in #580 before I accidentally remove the changes and it auto-closed; binding sites now create matching features in Genbank output

Also fixes another TFBS crash when incoming gene annotations are poor and mark genes small enough to be fully contained by a binding site. For an example, see attached:
image

The PR also adds a small tweak to glimmerhmm gene finding to prevent this kind of gene annotation, as it was the tool creating the poor quality gene annotations originally.


kblin (approved):
Looks good to me 👍

PR #586, opened 2023-04-19
Update mypy

SJShaw:
Updates mypy to 1.2.0 and changes options in `setup.cfg` to keep behaviour largely the same as previous versions.

Fixes one existing issues not previously detected and ignores a fairly reasonable case of false positives.


kblin (approved):
Looks good to me 👍

PR #592, opened 2023-04-27
Dependency updates and misc small fixes

SJShaw:
This PR handles a number of binary dependencies and related processing/issues:
- removes NRPSPredictor source as it's no longer used
- removes Java from `antismash.common.subprocessing` as no analyses remain with it as a dependency and it's expected that no new modules will depend on it
- adds support for `fimo` versions after 4.11.2 (tested with 5.5.2), but _**not**_ newer `meme` versions (the same 5.5.2 specifically ignores the `-nmotifs` argument)
- improves error reporting from `hmmscan` (though this doesn't fix `hmmscan`'s error being truncated, as that's not within antiSMASH), `diamond`, and `makeblastdb`

It also cleans up the style of `setup.py` and fixes an issue where reusing results from a run that had extra detection modules enabled and weren't enabled again in the next run would fail.


kblin (approved):
LGTM 👍

PR #593, opened 2023-04-28
chore: change version from future versions to 'dev'

SJShaw:
For versioning of development builds, the git hash is already included, so it is less confusing to have just `7.dev-<hash>` than a version which may not exist (e.g. `6.2.0-<hash>` as existed some time ago).


kblin (approved):
Looks good to me 👍

PR #598, opened 2023-05-04
pfam2go: Add warning on no pfam detection

zreitz:
I forgot the Pfam detection flag was `--clusterhmmer` and not `--pfam`. This autocompleted to `--pfam2go` which happily ran with only this understated debug-level message: `No Pfam domains found in record, cannot create Pfam to Gene Ontology mapping`

Enabling --pfam2go without a hmmer flag now gives the following error: `Pfam2go enabled without an Pfam analysis. The --clusterhmmer or --fullhmmer option is required.`

### DIFF
--- a/antismash/modules/pfam2go/__init__.py
+++ b/antismash/modules/pfam2go/__init__.py
@@ -26,11 +26,13 @@ def get_arguments() -> ModuleArgs:
     return args
 
 
-def check_options(_options: ConfigType) -> List[str]:
-    """ Checks options for conflicts.
-        No extra options, so they can't have conflicts.
-    """
-    return []
+def check_options(options: ConfigType) -> List[str]:
+    """ Pfam2go requires Pfams """
+    problems = []
+    if is_enabled(options) and not (options.fullhmmer or options.clusterhmmer):
+        problems.append("Pfam2go enabled without an Pfam analysis. "
+                        "The --clusterhmmer or --fullhmmer option is required.")
+    return problems
 
 
 def check_prereqs(_options: ConfigType) -> List[str]:


zreitz:
PS, I do think it would be nice to have `--pfam` as an option, for better agreement with `--tigrfam`.

kblin:
The reason we don't do this right now is that in principle, `pfam2go` works if you already have pfam annotations in the GenBank file in the right format, so you technically don't need to run `--fullhmmer` or `--clusterhmmer`.

kblin:
> PS, I do think it would be nice to have `--pfam` as an option, for better agreement with `--tigrfam`.

And that would then run `--clusterhmmer` or `--fullhmmer`?

zreitz:
> > PS, I do think it would be nice to have `--pfam` as an option, for better agreement with `--tigrfam`.
> 
> And that would then run `--clusterhmmer` or `--fullhmmer`?

`--clusterhmmer`, in full agreement with `--tigrfam`.



> The reason we don't do this right now is that in principle, `pfam2go` works if you already have pfam annotations in the GenBank file in the right format, so you technically don't need to run `--fullhmmer` or `--clusterhmmer`.

What's that format? I thought that was Catarina's big complaint, that the only trusted Pfams were from the `--reuse-results` json (which I did forget about here)

SJShaw:
> I forgot the Pfam detection flag was `--clusterhmmer` and not `--pfam`. This autocompleted to `--pfam2go` which happily ran with only this understated debug-level message: `No Pfam domains found in record, cannot create Pfam to Gene Ontology mapping`
> 
> Enabling --pfam2go without a hmmer flag now gives the following error: `Pfam2go enabled without an Pfam analysis. The --clusterhmmer or --fullhmmer option is required.`

The reason we originally went with the other message is that this version will fail on reusing results, e.g. 
```
$ antismash --clusterhmmer nisin.gbk
$ antismash --reuse nisin/nisin.json --pfam2go
Pfam2go enabled without an Pfam analysis. The --clusterhmmer or --fullhmmer option is required
```

We might want to instead add some kind of system to state dependencies, like a `get_dependencies()` that returns a list of modules, which the higher level parts of module running can then check for existing results and such.

PR #612, opened 2023-06-11
Fix: diamond version subcommand fails with --threads 

hyphaltip:
Only provide --threads to diamond command when it isn't a 'version' subcommand. 
The 'version' subcommand (breaks with v diamond 2.1.7) will  give these errors:
```
$ diamond version
diamond version 2.1.7

$ diamond version --threads 2
Error: Option is not permitted for this workflow: threads

$ diamond version --tmpdir /scratch/
Error: Option is not permitted for this workflow: tmpdir
```

### DIFF
--- a/antismash/common/subprocessing/diamond.py
+++ b/antismash/common/subprocessing/diamond.py
@@ -16,12 +16,14 @@ from .base import execute, get_config, RunResult
 
 
 def run_diamond(subcommand: str,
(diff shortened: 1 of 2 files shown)


kblin:
my bad, I forgot to update to the latest master branch before changing the patch

kblin:
I forgot to amend the function docs.

SJShaw (approved):

PR #620, opened 2023-06-27
html: swap from pyScss to libsass for CSS generation

SJShaw:
libsass has the equivalent function as pyScss, but is more reliable and has less warnings/issues with newer python. It does require a few changes to the SCSS file, but after those it builds a mostly identical (barring style and a couple of fixes) CSS file as a result.


kblin (approved):
Looks good to me 👍

PR #625, opened 2023-07-27
Misc small changes

SJShaw:
Changes:
- Adds a pair of options `--abort-on-invalid-records` (the default) and `--no-abort-on-invalid-records`. The default behaviour is unchanged, but the `no-abort` version will allow runs to simply discard invalid records and continue on those that remain valid
- Improves reporting of some input parsing failures where biopython itself crashes
- Renames the "NRPS/PKS monomers" sidepanel tab in the HTML output to "NRPS/PKS substrates", as only the unmodified substrates are shown
- Rewords the lanthipeptide rules to have consistent description style in order to simplify scraping of those files for uses other than antiSMASH
- Removes the `id` field from JSON output of records which weren't ever used and resulted in a great deal of `id="<unknown id>"` repetition bloating the file size

Fixes:
- Fixes lassopeptides being missed in the CompaRiPPson data generation script
- Fixes possible crashes in genefinding, where using record names for tempfiles could result in filenames being too long for the system (typically only applicable for cases where FASTA files embedded enormous amounts of information in the headers)
- Fixes a rare occurrence of cross-CDS NRPS/PKS modules also being cross-_region_ when the last CDS in a region and the first CDS in the next region were compatible
- Fixes a duplication in NRPS/PKS orderfinding that was only partially fixed in #530, specifically for cases where a single module with a termination domain was present in a CDS that also led with the tail of a cross-CDS module
- Fixes yet another broken string interpolation that I introduced in 8c37552
- Fixes a few typos in tests and tooltips

### DIFF
--- a/antismash/common/comparippson/data/build_data.py
+++ b/antismash/common/comparippson/data/build_data.py
@@ -126,7 +126,7 @@ def gather_entries(files: List[str]) -> List[Entry]:
                 # some RiPP modules keep motifs in different layouts
                 if name == "thiopeptides":
                     motifs = results["motifs"]
-                elif name in ["lanthipeptides", "sactipeptides", "thiopeptides"]:
+                elif name in ["lanthipeptides", "lassopeptides", "sactipeptides", "thiopeptides"]:
                     motifs = []
                     for _locus, motifs_for_locus in results["motifs"].items():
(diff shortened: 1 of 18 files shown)


kblin (approved):
Looks good to me 👍

PR #628, opened 2023-08-01
comparippson: work around mypy issue

SJShaw:
For some reason, on both Python 3.9 and 3.10 on _some_ systems, `mypy` (specifically 0.982, the pinned version) complains about missing attributes in the CompaRiPPson data script. Converting those imports/references from direct access (e.g. `antismash.modules.lassopeptides`) to `from antismash.modules import ...` avoids the issue.


kblin (approved):
Looks good to me 👍

PR #631, opened 2023-08-03
hmm_detection: return DUF692 to RiPP-like, since methanobactin is more specific

SJShaw:
When creating the methanobactin rule, which uses DUF692 *and* another profile, the DUF692 profile was removed, resulting in some previously detected RiPP-likes no longer being found.


kblin (approved):
Looks good to me 👍

zreitz:
Oops

PR #632, opened 2023-08-04
Update stable to 7.0.1

SJShaw:
Prepares for bugfix release 7.0.1, which fixes some issues identified in antiSMASH 7.0.0 and adds some small quality of life features:

* Fixes issues involving newer diamond versions (2.1 and above)
* Fixes some cross-CDS module behaviour in NRPS/PKS regions
* Fixes a possible crash when record names are extremely long
* Fixes a regression in the `RiPP-like` detection rule causing some clusters to be undetected
* Fixes lassopeptides being omitted from CompaRiPPson database generation
* Improves JSON output size by removing noisy unused fields
* Improves help output
* Improves reporting for some invalid input errors
* Improves miscellaneous descriptions (including detection rules, logging, and HTML output)
* Adds an option to drop invalid records and continue processing remaining records (`--no-abort-on-invalid-records`)

### DIFF
--- a/.github/workflows/pylint_issue_reporter.yml
+++ b/.github/workflows/pylint_issue_reporter.yml
@@ -5,6 +5,9 @@ on: [pull_request_target]
 jobs:
   lint_comparison:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
     steps:
     # install python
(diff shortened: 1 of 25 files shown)


kblin (approved):
Looks good to me 👍

PR #639, opened 2023-08-18
Small fixes and improvements

SJShaw:
A large collection of small things, some slightly more significant than others.

### Output:
- JSON results now include calculated clusterblast similarity scores, instead of users needing to derive it from the rest of the results
- JSON results for protoclusters now include the product category
- adds a link in the HTML download options to download the JSON results specifically
- CompaRiPPson now uses the same colouring for consensus sequences as the rest of that RiPP class does in reported sequences (mismatches between two colours will just be bolded, see below)
image
- lasso- and sactipeptides now only colour cysteines, as the others being coloured weren't really relevant to those classes
- the record identifier in HTML regions now explicitly uses the versioned identifier

### Tweaks:
- bumps the neighbourhood size of the `NI-siderophore` rule to fully contain BGC0000940 
- all generated gene translations for genes with alternate start codons now start with methionine 
- `bgc_seeds.hmm` now rebuilds if the `hmmdetails.txt` file is updated, due to issues with `wget` keeping older file modification timestamps, which led to some confusion as to why it wasn't updated
- PKS modules without a Trans-AT specific subtype in the ketosynthase are now considered complete if they *also* contain an AT-docking domain
- moved `RuleCategory` definitions and parsing from `detection.hmm_detection` to `common.rule_parser` as part of aiming for more easily specified alternate rulesets
- allows `antismash.common.subprocessing` functions to run without first running `antismash.config.build_config()`, using defaults for the system (for use as a library that needs to specify custom executable paths, `build_config(args...)` is still required)

### Fixes:
- fixed a regression causing NRPS/PKS modules with methylation domains to not correctly methylate the substrate
- removes some out of date restrictions in documentation for the rule parser

For future development, the following pylint messages are now considered failures rather than just warnings: `unused-variable` ,`comparison-with-itself`, and `self-assigning-variable`.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -6,6 +6,7 @@ disable=too-few-public-methods
 [MESSAGES CONTROL]
 # raising 'from' is almost entirely false positives
 disable=raise-missing-from
+fail-on=unused-variable,comparison-with-itself,self-assigning-variable
 
 [TYPECHECK]
 ignored-classes=Config,TestCase,Namespace
--- a/antismash/modules/nrps_pks/test/integration/integration_nrps_pks.py
+++ b/antismash/modules/nrps_pks/test/integration/integration_nrps_pks.py
@@ -11,9 +11,10 @@ from unittest.mock import patch
 from helperlibs.wrappers.io import TemporaryDirectory
 
 import antismash
-from antismash.common import path
+from antismash.common import path, secmet
 from antismash.common.test import helpers
 from antismash.config import build_config, destroy_config
+from antismash.detection import nrps_pks_domains
 from antismash.modules import nrps_pks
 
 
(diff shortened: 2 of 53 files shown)


antismash/modules/nrps_pks/test/integration/integration_nrps_pks.py
```diff
@@ -189,6 +189,7 @@ def test_substrate_methylation(self):
         # and the NRPS/PKS results for the candidate should have the right polymer
         region_results = results.region_predictions[region.get_region_number()]
         assert region_results[0].polymer == "(Me-Cys)"
+        assert region_results[0].smiles == "NC(C(C)S)C(=O)O"
```
kblin, line 192: I don't think this is the structure we should be generating.

.pylintrc
```diff
@@ -6,6 +6,7 @@ disable=too-few-public-methods
 [MESSAGES CONTROL]
 # raising 'from' is almost entirely false positives
 disable=raise-missing-from
+#fail-on=unused-variable,comparison-with-itself,self-assigning-variable
```
kblin, line 9: This looks commented out?

kblin (changes requested):
The main thing I'm tripping over is the commented-out pylint config changes, I think we need to revisit our methylation logic in general, unrelated of the current patch.

zreitz:
If you're going to extend the NIS neighborhood anyway, perhaps consider 10 kbp, which I use locally?

In the extreme, the rhodopetrobactin BGC extends 13.8 kbp from the IucA_IucC gene, including a biosynthetic gene at 11 kbp
[https://www.ncbi.nlm.nih.gov/nuccore/CP001096.1?report=genbank&from=2834779&to=2852134](https://www.ncbi.nlm.nih.gov/nuccore/CP001096.1?report=genbank&from=2834779&to=2852134) 
[https://pubmed.ncbi.nlm.nih.gov/29473283/](https://pubmed.ncbi.nlm.nih.gov/29473283/)

The avaroferrin BGC from Shewanella algae extends 10.6 kb 
[https://www.sciencedirect.com/science/article/pii/S2451945617301009](https://www.sciencedirect.com/science/article/pii/S2451945617301009)

antismash/modules/nrps_pks/test/integration/integration_nrps_pks.py
kblin: Let's delay that for a future fix

.pylintrc
SJShaw: It was, because I was testing it and forgot to uncomment it. It's fixed now.

SJShaw:
> If you're going to extend the NIS neighborhood anyway, perhaps consider 10 kbp, which I use locally?
> 
> In the extreme, the rhodopetrobactin BGC extends 13.8 kbp from the IucA_IucC gene, including a biosynthetic gene at 11 kbp https://www.ncbi.nlm.nih.gov/nuccore/CP001096.1?report=genbank&from=2834779&to=2852134 https://pubmed.ncbi.nlm.nih.gov/29473283/

I've changed it to be 14kb now, and added rhodopetrobactin as an explicit example in the rule.
```
RULE NI-siderophore
    CATEGORY other
    DESCRIPTION NRPS-independent IucA/IucC-like siderophores
    EXAMPLE NCBI CP001096.1 2834779-2852134 rhodopetrobactin  # https://doi.org/10.1111/1462-2920.14078
    CUTOFF 20
    NEIGHBOURHOOD 14  # for rhodopetrobactin
    CONDITIONS IucA_IucC
```

kblin (approved):
Looks good to me 👍

PR #642, opened 2023-08-25
chore: bump minimum python version to 3.9

### LINKED ISSUE #641: Installled v7.0.1 but "antismash --version" outputs 6.0.0
I've just downloaded the tar.gz source code for v7.0.1 and installed it following the full manual install instructions. Everything looks good when the installation ends, with the message ```Successfully installed antismash-7.0.1```. 

However, when I then run ```antismash --version``` the output is ```antiSMASH 6.0.0``` and running the program leads to an error related to not having the older muscle version installed, which I believe indicates it is version 6 being installed, not 7. 

Why is it installing version 6, not 7? How do I get the newest version if I have to do a manual install on my system?

I am working on a RHEL7 system.

SJShaw:
Somehow this was missing in the changes to 7.0, but we've been using Python 3.9 features for quite a while now.

Should fix #641

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -103,7 +103,7 @@ class PyTest(TestCommand):
 
 setup(
     name="antismash",
-    python_requires='>=3.7',
+    python_requires='>=3.9',
     version=read_version(),
     packages=find_packages(exclude=["*.tests", "*.tests.*", "tests.*", "tests"]),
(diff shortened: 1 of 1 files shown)


kblin (approved):
Looks good to me 👍

PR #643, opened 2023-08-25
Add extension options for detection rules

SJShaw:
Introduces new syntax for a detection rule section: `EXTENDERS <CDSCondition>`

The example use case is kirromycin, which is a TransAT-PKS. The current rule is
```
RULE transAT-PKS
    CATEGORY PKS
    DESCRIPTION Trans-AT polyketide
    EXAMPLE NCBI CP006259.1 382565-499230 kirromycin
    EXAMPLE NCBI CP001511.1 5000-30000 toblerol
    CUTOFF 45  # kirromycin (CP006259) has a range of ~42kb
    NEIGHBOURHOOD 20
    CONDITIONS cds(PKS_AT and not ANY_KS)
               and (
                cds(ATd and ANY_KS)
                # handle CP001511 5k-30k as per https://doi.org/10.1002/ange.201709056
                or cds(tra_KS and PP-binding)
               )
```
Previously, the cutoff for this rule had to be enlarged to 45kb to reach all the KS-containing CDSes. Some newer, more extreme cases were found where the distance from the last transAT module was 131kb away from the CDS with the AT domain. Making the cutoff that large would be problematic, so this alternative system was created to solve it.

By adding the line `EXTENDERS cds(ANY_KS and not PKS_AT)` to the rule (and lowering the cutoff to 20kb), this requires that the initial rule requirements are still necessary, but that the resulting protocluster's core location will now be iteratively extended for every CDS that satisfies `cds(ANY_KS and not PKS_AT)` within the cutoff distance from the initial core location. With this setup, both kirromycin and the 131kb long protoclusters are found correctly.

At the moment, these extension options are limited to a single CDS condition, as the extension check works only on a single CDS at a time (otherwise the existing rule style would work just as well). At some point in the future, this restriction could be lifted.

Another future possibility would be a similar `BLOCKERS` section, which would prevent any extension beyond any matching gene and could potentially also be applied to core rules. That's a discussion to be had at a later date.

### DIFF
--- a/antismash/common/hmm_rule_parser/cluster_prediction.py
+++ b/antismash/common/hmm_rule_parser/cluster_prediction.py
@@ -4,14 +4,19 @@
 """ Detects specific domains and defines clusters based on domains detected
 """
 
+import bisect
 from collections import defaultdict
 import dataclasses
 import logging
-from typing import Any, Dict, List, Optional, Set, Tuple
+from typing import Any, Dict, Iterable, Iterator, List, Optional, Set, Tuple
 
 from antismash.common import fasta, path, serialiser
 from antismash.common.hmmscan_refinement import HSP
 from antismash.common.secmet import Record, Protocluster, CDSFeature, FeatureLocation
+from antismash.common.secmet.locations import (
+    location_contains_other,
+    get_distance_between_locations,
+)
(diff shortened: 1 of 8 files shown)


kblin (approved):
Looks good to me 👍

PR #645, opened 2023-09-01
Small tweaks

SJShaw:
Both TigrFAM and ClusterHMMer used very silly ways of generating feature lists, these have been changed to use the functionality provided by the `Record` class. The results aren't affected.

As the JSON results serialiser changes introduced in c2d08646 were just a small addition that regenerate, the compatible schema list has been bumped. This won't likely have much affect on most use cases of actual results reuse as rulesets have changed, but there are some edge cases where it's useful.

The SMILES for the `cap` substrate have been rearranged so the backbone ordering is consistent with other substrates.


kblin (approved):
Looks good to me 👍

PR #646, opened 2023-09-01
clusterblast: remove raw diamond output files

SJShaw:
These are generally misleading for people, who expect these to be the output of the analysis and not the raw hits. As such, it would be better if they were removed, especially given the file sizes that can be reached.


kblin (approved):
Looks good to me 👍

PR #649, opened 2023-09-07
HTML generation optimisation

SJShaw:
Around 25% of the time spent in HTML generation, including when using `--reuse`, was spent calculating the GC content of records. This was largely due to repeatedly calculating it in a loop over regions within the record within the HTML template that generates each region view.

Even after moving the repetition out of the loop, it's still a repeated calculation, as the TTA module will use it along with the HTML output module. To fix this, `Record` now caches the value after it's been calculated once.

To go even further, this doesn't even need to be calculated at all when using `--reuse` if the value is kept in the JSON results. That's fixed by adding a `gc_content` field to the JSON records, and the required infrastructure to pass that through to `Record`'s init and also `from_biopython()`.

As a last bonus, the HTML and SVG generation time is now included in the debug logging of module timings.

### DIFF
--- a/antismash/common/secmet/record.py
+++ b/antismash/common/secmet/record.py
@@ -78,7 +78,7 @@ class Record:
                  ]
 
     def __init__(self, seq: Union[Seq, str] = "", *,
-                 transl_table: int = 1,
+                 transl_table: int = 1, gc_content: float = -1.,
                  **kwargs: Any,
                  ) -> None:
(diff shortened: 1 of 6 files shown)


kblin (approved):
Looks good to me 👍

kblin:
not sure what happened with codecov here, we might want to rebase it after merging #648 and try again

PR #652, opened 2023-09-13
Misc fixes

SJShaw:
The final log message for a run that failed due to invalid inputs was outside the context in which logging was set up, this meant that log files never received this last, rather important, message.

I've added three more small fixes to this PR:
- `jsonschema` version is now pinned until some deprecations are handled
- uses of `matplotlib.pyplot` have been changed to be more agnostic of packages versions (including python versions)
- works around a bug in some versions of diamond

The diamond bug is that the subcommand `makedb` considered the use of  `--tmpdir` to be an error, and it affected versions 2.1.0 to 2.1.6. Handling has been added to ensure that `run_diamond_makedb()` avoids that parameter when those versions are in use.  This is particularly annoying in Debian 12/bookworm, for which the current diamond package is 2.1.3.

### DIFF
--- a/antismash/__main__.py
+++ b/antismash/__main__.py
@@ -4,7 +4,6 @@
 # A copy of GNU AGPL v3 should have been included in this software package in LICENSE.txt.
 """Run the antiSMASH pipeline"""
 
-import logging
 import os
 import sys
 from typing import List, Optional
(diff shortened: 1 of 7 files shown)


kblin (approved):
Looks good to me 👍

PR #659, opened 2023-10-19
common.all_orfs: fix minimum length not being respected in all cases

SJShaw:
Fixes an issue where finding ORFs smaller than the default minimum didn't respect the minimum provided in internal functions, which led to ORFs smaller than 60 nucleotides being missed.


kblin (approved):
Looks good to me 👍

PR #661, opened 2023-10-23
Tweaks for ClusterCompare

SJShaw:
Two issues related to subregions caused poor results in ClusterCompare when comparing against MIBiG hits:

1. CDS features with hits to rule detection profiles were not annotated when only in a sideloaded subregion and not a protocluster. This was due to sideloaded annotations being added _after_ the detection ran. This caused ClusterCompare to be missing components when the region in question had a subregion (e.g. all of the MIBiG antiSMASH-generated pages).
The fix was moving the sideloaded detection to the "full genome" stage prior to the "area formation stage".

2. The "Protocluster to Region" view for the ClusterCompare visualisation deliberately omits subregions since they aren't protoclusters. The "Region to Region" view is more accurate in that case. Since all antiSMASH-generated MIBiG pages have subregions, this could cause some confusion.
The fix for that is making MIBiG-specific comparisons default to "Region to Region" instead of "Protocluster to Region".

### DIFF
--- a/antismash/detection/sideloader/__init__.py
+++ b/antismash/detection/sideloader/__init__.py
@@ -20,7 +20,7 @@ from .html_output import generate_html, will_handle
 
 NAME = "sideloader"
 SHORT_DESCRIPTION = "Side-loaded annotations"
-DETECTION_STAGE = DetectionStage.AREA_FORMATION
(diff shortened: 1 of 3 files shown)


kblin (approved):
Looks good to me 👍

PR #662, opened 2023-10-24
Misc fixes

SJShaw:
A collection of small fixes, the important ones being:
- NCBI genomic context links are no longer present by default, since they weren't relevant for many inputs. For them to appear they need to be explicitly enabled with `--html-ncbi-context`.
- Lasspeptide precursors that were in an overlapping neighbourhood of two separate lassopeptide protoclusters no longer crash with a duplicate name

The remainder are small changes for documentation, lint errors, and making some non-deterministic multiprocess testing more reliable.

### DIFF
--- a/antismash/common/hmm_rule_parser/cluster_prediction.py
+++ b/antismash/common/hmm_rule_parser/cluster_prediction.py
@@ -551,14 +551,10 @@ def create_rules(rule_files: list[str], signature_names: Set[str],
                  ) -> List[rule_parser.DetectionRule]:
     """ Creates DetectionRule instances from the default rules file
 
-        Updates existing_aliases with any aliases found.
-
         Args:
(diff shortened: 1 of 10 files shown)


SJShaw:
> Have you already enabled context for the webserver when using "Get from NCBI"?

No, these changes won't be applied to the (non-dev) webserver until we have at least a minor version release, at which point we'll update the webservice UI at the same time as deploying the release.

It'll be automatically enabled if the option to fetch an accession from NCBI is used, and we haven't yet looked into whether it should have an extra toggle option on the UI for when users supply their own input files.

zreitz:
Ah right, forgot about the delay between the two versions. Thanks! 

IMO an extra toggle seems too cluttery. Can you use the same regex pattern that you use for NCBI validation to make an educated guess?

kblin (approved):
LGTM 👍

PR #671, opened 2023-11-08
lanthipeptide: avoid hitting too long a gene even if it has a precursor domain

SJShaw:
Fixes rare cases of genes that merge precursors and LANC-likes which resulted in prepeptide cores of 1000+ aminos.


kblin (approved):
Looks good to me 👍

PR #680, opened 2023-12-04
Categorization for Flavin-dependent monohalogenases accepting tryptophan

JudSze:
This is the initialization for the categorization of halogenating enzymes.
Right now it contains the categorization workflow for Flavin-dependent monohalogenases accepting tryptophan substrate.
There are two separate pHMMs. One for Tryptophan(Trp)-5 halogenases and one for Tryptophan-6/7 halogenases.
A low-quality and a high-quality cutoff is defined for the Trp-5 halogenases to pick up enzymes which are special in that group.
Only one cutoff is defined for the Trp-6/7 profile.

The categorization happens by:
* doing hmmsearch against the pHMMs
* checking whether the hit meets the predefined cutoffs
* in case of tryptophan-5 halogenases we check the signature residues
* to differentiate between tryptophan-6 and tryptophan-7 halogenases we use signature residues as well

If there is hit in both profiles, we choose the most likely outcome by comparing the confidence of the categorization.
E.g. if there is a hit to the Tryptophan-5 profile, the signature residues match, but it only meets the low-quality cutoff, the confidence is less (e.g. 0.5) compared to when it has a hit to the Trp-6/7 profile with the higher cutoff.

The information of the pHMM profiles is bundled into HmmSignature class. This includes the cutoff.
The information of the signature residues and their positions in the pHMMs are available in the FlavinDependents class.

The hits to the profiles are stored in the HalogenaseHmmResult class.
If the hit is a match to the trp-5 or trp-6/7 groups, then the information is stored in the Match class.
The Match instances are also available in the potential_matches attribute of the HalofenaseResult class.

The final categorization of an enzyme is accessible through a HalogenaseResult instance.
If there are several matches the substrate-and regio-specificity of the enzyme is decided by comparing the confidences of the matches. In case there are several matches with the same confidence, the substrate-and regiospecificity is not defined.

The function which does the whole workflow is called specific_analyses and takes a record instance which holds information of the identified cluster.


JudSze:
Reopening, because the FDH family **detection** part is complete

kblin (dismissed):
Looks good to me 👍

SJShaw (approved):

kblin:
Merged with a734a06, but GitHub somehow still thinks this PR is not merged.

PR #685, opened 2023-12-22
common: ensure that pressed HMM database files are not empty

zreitz:
If `hmmpress` is terminated early, the broken database will pass `ensure_database_pressed` because all four `.h3x` files are immediately created. This leads to a `RuntimeError` during analysis.

Now, `ensure_database_pressed` checks that `h3x` file size is greater than zero.


SJShaw (approved):
I removed the extra generator construction, but otherwise it looks good to me.

PR #688, opened 2024-01-03
Add a detection rule for 7-hydroxytropolone

zreitz:
* Key reference: [10.1128/jb.00087-18](https://doi.org/10.1128/jb.00087-18)
* Although the exact biosynthetic pathway is unknown, four genes are required via knockout ([10.1128/jb.01756-14](https://doi.org/10.1128/jb.00087-18)). This rule also requires all four, using the relevant Pfam domains.
* Most pHMMs were already present. This rule adds one more, `PF13279` (4HBT_2).
* The rule is named `hydroxytropolone` instead of `7-hydroxytropolone` because `secmet.scss` didn't work, presumably due to starting with a number.
* Because 7-HT was previously shown to function as a siderophore ([10.1007/s10534-016-9954-0](https://doi.org/10.1128/jb.00087-18)), I chose to use the metallophore coloring.
* The rule was tested on Pseudomonas strains with and without the BGC. By BLAST and CAGECAT, it doesn't look like any non-Pseudomonads have all four genes in a cluster.


SJShaw:
Regarding the naming, CSS doesn't allow leading digits, but if you use the character code instead of a plain number in the (S)CSS (i.e. `.\37-hydroxytropolone` instead of `.7-hydroxytropolone`) , it will work.

kblin (approved):
Looks good to me 👍

PR #696, opened 2024-02-14
Misc fixes

SJShaw:
- Wraps a few user input errors with more graceful exits (i.e. no backtraces)
- Adds more information to bad sideloaded area inputs
- Changes ClusterCompare to not multi-protocluster normalisation function when only a single protocluster is involved
- Adds a workaround for `makeblastdb` failing on user input with long or incompatible locus tags/identifiers
- Adds a record identifier length check in case the identifier is longer than file system can support in output files
- Allows merging of T1PKS cross-CDS modules where the head contains only the KS and the tail would otherwise be a valid starter module by itself
- Fixes reporting of HTML time taken by fixing an operator precedence bug when averaging times over multiple records


kblin (approved):
Looks good to me 👍

PR #698, opened 2024-02-22
Enforce ignored files not being present in pull requests

SJShaw:
Adds a pull request check (trivially extendable to include pushes as well) that finds and lists any files that match ignore filters.

To avoid bypassing this (intentionally or accidentally) via first modifying `.gitignore`, the default branch's `.gitignore` will be used over the one in the pull request itself.

PR #699, opened 2024-02-22
chore: extend git ignore exceptions for known cases

SJShaw:
A prerequisite for #698, since the action it's introducing is clever enough to undo these changes when in the same PR.


kblin (approved):
Looks good to me 👍

PR #700, opened 2024-02-23
Reduce size of datafiles for HTML/javascript

SJShaw:
`regions.js` is used by the javascript in the HTML output to provide all the data for dynamic interaction.

This PR makes two trivial changes to whitespace formatting in the contents of that file. For the purposes of size comparison, a run of the master branch over NC_003888.3 with all options on is 9.4Mb. 

CDS details panels are pre-rendered, but the templating didn't use jinja's whitespace reduction markers. The removal of this extraneous whitespace reduced the file size to about 95% (to 8.9Mb) of the example.

The file itself contained the JSON objects with human-friendly indents. Given the depth of these objects and their children, removing those indents reduced the file size further, to about 82% (7.8Mb) of the example.

There will be more HTML templates through the codebase that can have unnecessary whitespace removed, but the CDS panels are the most numerous by far, making it the bulk of the possible savings. At some point this should be shifted to a less static approach, but that's a future PR.


kblin (approved):
Looks good to me 👍

PR #702, opened 2024-02-28
Polyhalogenated pyrrole rule

JudSze:
Add a role for polyhalogenated pyrrole


kblin:
I think the rule should go into the strict rule-set.
Our explanation for what the relaxed rules are is
> Detects partial clusters missing one or more functional parts.

and that rule contains all the functional parts to create polyhalogenated pyrroles.

SJShaw (approved):
Moved to strict set

PR #704, opened 2024-03-20
Update biopython dependency to 1.81

SJShaw:
1.81 was chosen because 1.83 introduces a whole heap more issues than it solved. 1.81 does allow use of an updated GFF parser.

Now that biopython has a shared base class for locations, the opportunity was taken to add a mixin/wrapper for the location classes to add most of the static location functions as instance methods.

### DIFF
--- a/antismash/common/all_orfs.py
+++ b/antismash/common/all_orfs.py
@@ -11,10 +11,9 @@
 
 from typing import Iterable, List, Optional, Tuple
 
-from Bio.SeqFeature import FeatureLocation
-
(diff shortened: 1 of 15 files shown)


SJShaw:
There's a few bits of improvement that are possible here, so I'll mark it as a draft for the moment until I get to them. For example the `Location.is_bridged()` method is ambiguously named and inconsistent with the naming used in #697 for similar purposes.

SJShaw:
Largely superseded by #710, the location subclassing will be split into a different PR that will follow later.

PR #707, opened 2024-04-08
build: convert setup.py based setup to pyproject.toml

afuetterer:
Hi, this is my first PR to the project. Please feel free to close, if the changes are not desired.

This PR proposes the following changes:

- remove `setup.py` based project setup
- use modern python project setup with `pyproject.toml`
- move as much tooling configuration into pyproject.toml (flake8 and pycodestyle still require the `setup.cfg`)

Please let me know what you think.

Refs:
- https://packaging.python.org/en/latest/tutorials/packaging-projects/


afuetterer:
Should I close this?

kblin:
The reason we haven't responded on this one yet is that I'm not sure we want to go with hatchling for our build system, and we didn't have time to discuss this properly yet.

afuetterer:
Alright, no problem. Thanks for the feedback.

Hatchling is just one backend that uses the toml file.
Setuptools, flit and pdm are other available options, see https://packaging.python.org/en/latest/tutorials/packaging-projects/#choosing-a-build-backend

I can change to setuptools, for example.

Do you want me to keep this open or is this change not desired (at the moment)?

PR #710, opened 2024-04-11
Dependency version upgrades

SJShaw:
- Bumps minimum python version from 3.9 to 3.11 to allow for new features which simplify future development (e.g. the `Self` type).
- Bumps `mypy` from 0.982 to 1.9.0, along with associated fixes for existing types
- Bumps `biopython` from 1.78 to 1.81, along with associated fixes
- Bumps `bcbio-gff` from 0.70 to 0.71 (which required a biopython newer than 1.78), fixing some issues with GFF inputs


kblin (approved):
Looks good to me 👍

PR #716, opened 2024-05-16
secmet: fix CDS feature name generation breaking CSS/HTML output

SJShaw:
In cases where a CDS feature had no identifiers (`locus_tag`, `gene`, `protein_id`), names are autogenerated using the location. When the location had ambiguous positions, the resulting name included the string representation of that position, e.g. `cds<123_456`, which causes issues in HTML/CSS.

This change removes the ambiguous marker from the name, resulting in `cds123_456`. This may collide with other unidentified CDS features at the same coordinates without ambiguous positions, but that is an issue best solved by correcting the annotations before they hit antiSMASH.


kblin (approved):
Looks good to me 👍

PR #719, opened 2024-05-22
common.json: fix JSON conversion of subclasses of stdlib types

SJShaw:
The stdlib python does not respect subclasses of types without either a custom encoder being passed to every json.dump(s) call or forcing pure-python conversion with by setting `json.encoder.c_make_encoder` to False. This caused the JSON helper classes in `antismash.common.json` to be converted to an empty dictionary when using the C implementation, which in turn prevented the NRPS/PKS domain view from showing anything after the changes in 09e7076bc where indentation was removed to save disk space.

Both options are poor because it's easy to miss the extra option or it becomes much slower to convert to JSON for the large HTML JSON dumps.

As a solution, all json library calls have been swapped to use `antismash.common.json` instead, taking advantage of the opportunity to switch to the much faster `orjson` library instead of the standard library.

This comes at the cost of slight flexibility, as indent levels in `orjson` are either 2 spaces or no indentation, but that seems bearable, given the speed increase.

Along with speed increases, the changes also add more flexibility in class conversion. All classes with a `to_json()` or `__json__()` method will now convert automatically, which will simplify many instances of classes with attributes of other classes needing conversion to JSON. This only applies to conversion *to* JSON, converting *from* JSON must still be specified.

### DIFF
--- a/antismash/common/comparippson/data/build_data.py
+++ b/antismash/common/comparippson/data/build_data.py
@@ -7,10 +7,11 @@
 import argparse
 from dataclasses import dataclass
 import glob
-import json
 import os
 import sys
 from typing import Any, Dict, List
+
+from antismash.common import json
 
 
 @dataclass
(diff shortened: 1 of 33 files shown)


kblin (approved):
Looks good to me 👍

PR #720, opened 2024-05-28
Add CAL domain as a loader

zreitz:
Many lipopeptide siderophore NRPSs start with a fatty acid adenylation domain belonging to the CAL domain family (ex: [pyoverdine's PvdL](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3220798/)). Because `CAL_domain` isn't in the list of loaders, the module isn't recognized as valid. Here's PvdL:
image

In this more unusual case (NZ_ABLD01000016), the gene with the true initiation module is excluded from the modules tab, leading to another gene to incorrectly be labeled as the "complete" starter module.
image
image

This pull request adds the CAL domain to the list of loaders, which then allows the `CAL_domain - PCP` module to be recognized as complete:
image

BGC0000180 is an example of a non-siderophore BGC that will be affected (correctly, I'd say).

### DIFF
--- a/antismash/detection/nrps_pks_domains/module_identification.py
+++ b/antismash/detection/nrps_pks_domains/module_identification.py
@@ -164,6 +164,10 @@ class Component:
         """ Returns True if the component can function as an acyltransferase domain """
         return self.label in ACYLTRANSFERASES
 
+    def is_cal(self) -> bool:
+        """ Returns True if the component can function as an acyltransferase domain """
+        return self.label == "CAL_domain"
+
     def is_condensation(self) -> bool:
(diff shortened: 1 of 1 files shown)


kblin (approved):
Looks good to me 👍

SJShaw:
I have a [related branch](https://github.com/SJShaw/antismash/tree/AHBA_rule) that covered this change, along with the SMILES for the CAL domain predictions. It also added a rule for the AHBA prediction, which we probably don't want, but I think we ought to bundle the further handling of the module at the same time as we start treating the modules as being complete.

SJShaw:
Closing in favour of the now-merged #724

PR #723, opened 2024-07-04
Small fixes and improvements

SJShaw:
A collection of small changes, most of which are simple fixes or improvements. The remainder are aimed at supporting future use cases or use as a library (e.g. adding `subprocessing.prodigal` runners.

- Improves some testing helpers, and performance for various areas.
- Improves communication of `--sideload-by-cds` with a name that does not exist in inputs. As no mechanism exists (yet) for checking all records with all options, for now this just logs a warning.
- Adds explicit tracking of `source` features in records, for future use in multi-source inputs.
- Fixes an issue in lassopeptide scoring where checking that another component was within a certain distance would never fail.
- Fixes an issue with some very specific NRPS/PKS module constructs resulting in crashes due to unexpected domain ordering when fetching the module's protein location.

### DIFF
--- a/antismash/common/secmet/__init__.py
+++ b/antismash/common/secmet/__init__.py
@@ -22,6 +22,7 @@ from .features import (
     Prepeptide,
     Protocluster,
     Region,
+    Source,
     SubRegion,
 )
 from .qualifiers import GeneFunction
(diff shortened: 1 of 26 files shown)


kblin (approved):
Looks good to me 👍

PR #724, opened 2024-07-08
Add support for treating modules with CAL domains as complete, including monomer/polymer generation

SJShaw:
A more complete version of #720. 

- adds SMILES for all predictions made for what's loaded by CAL domains
- treats CAL domains and starter domains and loader domains so that the module itself can count as a starter module
- adds support for monomer modification if dehydrogenase domains are present in the module (the only affected substrate is `AHBA`)
- treats modules with CAL domains as PKS domains to both include the monomer in the predicted polymer and enable non-colinear ordering 

Using some of the examples from the other PR, here's the results:
image

image

image

And an unmodified `AHBA`, non-colinear example from NZ_MEAX01000018:
image


kblin (approved):
Looks good to me 👍

PR #729, opened 2024-08-08
hmm_detection: extend and improve the mycosporine rule

JudSze:
Extend mycosporine rule to detect ligases participating in mycosporine-like amino acid biosynthesis and add the paper the rule relies on (https://journals.asm.org/doi/10.1128/aem.00727-14)


kblin (approved):
Looks good to me 👍

PR #733, opened 2024-08-16
Update ClusterCompare with modules that use CAL domains 

SJShaw:
This fixes a crash that occurred when a ClusterCompare query region included a module that used a CAL domain (as introduced in #724), because they were registered as a having a module type of "unknown" in the features.
This has been corrected by adding a new CAL-specific module type in `secmet` and by setting that type in `nrps_pks_domains`.

Adding on that fix, ClusterCompare has been changed from a binary weighting system (i.e. PKS and NRPS) for modules to a system that can take an arbitrary number of types. Initially, these CAL domains have been included in a "generic domains" category, intended for those modules which are too constrained/specific for their own category. This may need to be updated in future, if detection can find a larger variety of these modules.

The building of ClusterCompare databases is unaffected, as that uses `secmet` module types already.

### DIFF
--- a/antismash/common/secmet/features/module.py
+++ b/antismash/common/secmet/features/module.py
@@ -20,6 +20,7 @@ class ModuleType(Enum):
     UNKNOWN = 0
     NRPS = 1
     PKS = 2
+    CAL = 3  # this is very specific to CAL domains, but they're common enough to justify this
 
     def __str__(self) -> str:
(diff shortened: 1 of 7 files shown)


kblin (approved):
Looks good to me 👍

kblin (approved):
Looks good to me 👍

Issue #732, opened 2024-08-15
Creating new rules


marlaux:
Dear all
I'm trying to create a customized rule and I'm getting the error 
ERROR    15/08 16:47:58   Incompatible options detected:
Rules contained identifers without signatures: PF16861

Could you please help me to understand what I am missing?

My rule:
RULE saxitoxin
    CATEGORY RiPP
    DESCRIPTION Saxitoxin cluster, core proteins
    # See https://doi.org/10.1128/AEM.00353-08
    EXAMPLE NCBI DQ787200.1 1-37606 saxitoxin # BGC0000188
    RELATED FA_hydroxylase, Sulfotransfer_3, Sulfotransfer_5, RMNT_CmcI, APS_kinase, YqcI_YcgG, FG-GAP_3, DUF5992, MatE, PhoU
    CUTOFF 30
    NEIGHBOURHOOD 30
    CONDITIONS PhyH and cds(Carbam_trans_N and PF16861) # SxtS and SxtI Carbam_trans_C-cyano 
                and cds(Aminotran_1_2 and (PP-binding or PF04055)) # SxtA Radical_SAM
                and cds(PF00383 and PF08211) # SxtB dCMP_cyt_deam_1-cyano and dCMP_cyt_deam_2-cyano
                and cds(PF00355 and PF19112) # SxtHT Rieske and VanA_C-cyano
                and cds(PF13187 or PF12838) # SxtW Fer4_9-cyano or Fer4_7-cyano
                and cds(PF00890 or PF02910) # SxtV FAD_binding-cyano or Succ_DH_flav_C-cyano
                and adh_short and PF19588 and PF19451 # SxtU and SxtJ and SxtK DUF5989-cyano sxtJ-cyano

My edited hmmdetails.txt:
... line 368
PF16861 	Carbamoyltransferase C-terminus	30	PF16861.hmm
PF00355	Rieske [2Fe-2S] domain	30	PF00355.hmm
PF00383	Cytidine and deoxycytidylate deaminase zinc-binding region	30	PF00383.hmm
PF08211	Cytidine and deoxycytidylate deaminase zinc-binding region	30	PF08211.hmm
PF00890	FAD binding domain	30	PF00890.hmm
PF02910	Fumarate reductase flavoprotein C-term	30	PF02910.hmm
PF12838	4Fe-4S dicluster domain	30	PF12838.hmm
PF13187	4Fe-4S dicluster domain	30	PF13187.hmm
[...]

PR #734, opened 2024-08-20
common.signature: forbid leading and trailing whitespace in signature identifiers

SJShaw:
Prevents difficult to find issues in whitespace splitting in HMM details files, e.g. #732.

At some point in the future, a clearer file format would be much better, but for the moment this is the simplest fix.


kblin (approved):
Looks good to me 👍

PR #741, opened 2024-10-02
Added RBS finding to get more confidence on RiPP precursors

FriederikeBiermann:
Enables future check for ribosomal binding site upstream of RiPP precursor peptides

### DIFF
--- a/antismash/common/all_orfs.py
+++ b/antismash/common/all_orfs.py
@@ -19,7 +19,7 @@ from antismash.common.secmet.features import CDSCollection
 START_CODONS = ('ATG', 'GTG', 'TTG')
 STOP_CODONS = ('TAA', 'TAG', 'TGA')
 RIBOSOMAL_BINDING_SITE = re.compile('?:GGA[AG]G')
-FLEXIBLE_RIBOSOMAL_BINDING_SITE = re.compile('(?=(?:[GA][^GA]{0,2}){6})')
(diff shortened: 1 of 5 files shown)


SJShaw:
There's a few issues with this implementation.

It only affects those features found with `all_orfs`, so it doesn't do anything for those precursors are already annotated in the inputs to antiSMASH or found with genefinding, if the user requests it.

The method used to set the `rbs` attribute on `CDSFeature` will crash when attempting to access it for any instance created without it.
E.g.:
```python
cds = CDSFeature(location, translation, locus_tag="test")
cds.rbs
...
AttributeError: 'CDSFeature' object has no attribute 'rbs'

cds = CDSFeature(location, translation, locus_tag="test", rbs=True)
cds.rbs
(True)
```

There isn't any output handling for this value, so it won't be annotated in either Genbank output or JSON output.

I don't feel that this is relevant to store in `CDSFeature`, as it is. I think this would be best kept in a post-genefinding step, such as the RiPP modules that are predicting the precurors to begin with.

FriederikeBiermann:
I now removed the RBS from the cds_feature and made the has_RBS function a more general function that can be used by the RiPP modules. I do not think it should be used on genes detected by more advanced gene calling, because those tools usually already check for the presence of an RBS.

kblin:
Now that the circular genome change has landed, this PR will need updating for the new way of handling how to get up- and downstream coordinates. But maybe I'd also put this on hold until we have a piece of code that actually relies on the RBS detection?

PR #743, opened 2024-10-03
Fix extreme slowdowns when running tests on machines with many cores

SJShaw:
When using `build_config(..., isolated=True)`, the intent is to not respect config files. However, when running on a machine with many cores (e.g. 256), the tests that use this isolated mode run much, much slower due to contention, i.e. 5-10 times longer than they should.

By changing the isolated mode to also keep the CPU count, this slowdown can be avoided, provided the config file is set.

When running on the command line, this still requires that the user supply, either with a config file or with the relevant argument, the number of CPUs desired.


kblin (approved):
Looks good to me 👍

PR #748, opened 2024-10-11
Updated CDPS rules

lisavader:
I added a new rule called 'fungal_CDPS' to detect fungal cyclodipeptide synthases that do not show sequence similarity to bacterial CDPSs. The exact biochemical function of these enzymes is still unknown.
This covers MIBiG entries BGC0001989 and BGC0002321.

This pull request contains:
- New hmm profile 'CDPS_fung'.
Accessions used for building the profile: XP_453056.1, QBM88528.1, XP_037144654.1, CDO92882.1, XP_022393384.1, XP_008599534.1, XP_003069960.2
Two of these sequences encode characterised enzymes involved in pulcherrimin biosynthesis: XP_453056.1 (paper: https://doi.org/10.1073/pnas.1806268115) and QBM88528.1 (paper: https://doi.org/10.1111/mmi.14272). The other five sequences have not been experimentally verified but are neighboured by genes similar to those found in the pulcherrimin BGC (encoding for a p450, MFS transporter and transcription factor). This is also described in https://doi.org/10.1073/pnas.1806268115.

- A new rule called 'fungal_CDPS', for which the only condition is a hit tot the 'CDPS_fung' hmm with a bitscore of at least 100.
The bitscore was determined by running the profile against a set of 313 fungal genomes and manually inspecting the output. Leave-one-out validation yielded a minimum bitscore of 154.7 and a maximum bitscore of 605.1.

- Removal of the hmm profile 'cycdipepsynth'
All regions in the antismash database that contain core genes with hits to 'cycdipepsynth' profile also have hits to the newer and more general 'CDPS' profile for bacterial CDPSs. None of the regions are uniquely identified by 'cycdipepsynth'. The profile does not represent a subclass within CDPSs. Therefore, this profile is redundant and will be removed.


kblin (approved):
Looks good to me 👍

PR #739, opened 2024-09-18
Support the atropopeptide class of RiPPs

kblin:

PR #750, opened 2024-10-16
common.hmm_rule_parser: allow dynamic profiles access to existing HMMer profile hits

SJShaw:
Changes dynamic profiles to pass in hits from existing HMMs from earlier in detection. This allows for dynamic profiles to avoid duplicating any work in finding the broader applicability before moving on to their more specific checks.

Should replace the base commit in #739, which reintroduced obsolete code and styling.


kblin (approved):
LGTM 👍

PR #751, opened 2024-10-17
Overview rearrangement

SJShaw:
Changes the most similar known cluster section of the HTML overview which was causing frequent misinterpretations.
- anything with an old similarity of less than 15% is no longer shown
- quantitative values have been changed to qualitative
- colouring has been made consistent regardless of row striping, as that was also misinterpreted to have some significance

Before:
image

After:
image

Also fixes an issue where the clusterblast module didn't regenerate it's HTML content including that in the overview.


kblin (approved):
Looks good to me 👍

zreitz:
Note: as currently implemented,
> anything with an old similarity of less than **or equal to** 15% is no longer shown

PR #756, opened 2024-11-04
hmm_detection: Exclude hits to Lys2 in NRPS and NRPS-like rules

lisavader:
Lys2 is part of the α-aminoadipate pathway for lysine biosynthesis in fungi.
Although the architecture of this enzyme fits the NRPS / NRPS-like rules, it shouldn't be picked up as it is part of primary metabolism.
TIGR03443 spans the entirety of this enzyme and can be used to reliably detect Lys2 homologs.

This pull request contains:
- Add hmm and add hmm details for TIGR03443
- Change rules NRPS and NRPS-like to exclude genes that hit TIGR03443 (alpha_am_amid)
- Add new loose rule 'lysine' for which the only condition is a hit to TIGR03443


kblin (approved):
LGTM 👍

PR #757, opened 2024-11-06
clusterblast: fix storing large, unfiltered dictionaries

SJShaw:
Fixes an issue I introduced in #701. The original change was to keep reference gene information available for the purposes of drawing the clusterblast SVGs dynamically. The problem was that it made a copy of the reference protein dictionary for each region, without any filtering.

This caused some systems to run out of memory and kill the antiSMASH job, with up to 40GB used before that happened. On my machine, running a large input of 130 contigs, memory use hit 20GB (for the whole antiSMASH run, not just these dictionaries) in 5 minutes without getting through the vast majority of those contigs.

After filtering out all reference proteins unrelated to the region matches, the memory use for all 130 contigs came down to 14GB (taking about an hour on my machine, for comparison). Filtering the set further to only include those being displayed in the HTML took the memory use down even further, to a little over 6GB (again, the whole antiSMASH run over those 130 decently sized contigs).

Also refactors the fetching of the display limit a tiny bit to make testing a little less obnoxious.

### DIFF
--- a/antismash/modules/clusterblast/html_output.py
+++ b/antismash/modules/clusterblast/html_output.py
@@ -381,7 +381,7 @@ def generate_javascript_data(record: Record, region: Region, results: ClusterBla
             for query, reference in score.scored_pairings:
                 pairs_per_ref[reference.name].append((query, reference))
             assert pairs_per_ref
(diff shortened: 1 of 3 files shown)


kblin (approved):
Looks good to me 👍

PR #760, opened 2024-11-26
cluster_compare: penalise single gene hits to multi-gene references

SJShaw:
ClusterCompare only adds an ordering metric when multiple genes in the reference have hits. This raises some single gene hits in the ranking more than they deserve, due to the final ranking using a geometric average of the metrics. This caused rankings like these (the top three hits from a lanthipeptide in NZ_BEWD01000001.1):
image

By adding a penalty for these, as if there was a second segment, the scores for these kinds of hits are reduced, leading to the top three changing to these:
image

The final score of venezuelin, since it's used in the example, changes from the initial 0.61 to 0.52, below the lowest of the multi-gene hits against similar class II lanthipeptides within MIBiG. This is true for other single-gene hits in the same example, though they're not shown in the screenshots.

### DIFF
--- a/antismash/modules/cluster_compare/data_structures.py
+++ b/antismash/modules/cluster_compare/data_structures.py
@@ -408,9 +408,7 @@ class ReferenceScorer:
     def final_score(self) -> float:
         """ Calculates the singular score of a result from its metric scores """
         if self._final_score is None:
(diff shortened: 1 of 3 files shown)


kblin (approved):
LGTM 👍

PR #761, opened 2024-11-26
NRPS/PKS HTML updates, including link out to PARAS

SJShaw:
Adds links out to PARAS as a downstream adenylation domain predictor.

This required moving the domain drawing generator from `detection.nrps_pks_domains` into `html.visualisers`, in order to more easily access results from multiple modules. A side benefit of this change is that the last special case in the output JS structures is now removed, cleaning up some of both the HTML generation and the JS data handling.

Also takes the opportunity to fix both the NORINE links out referring to old/dead URL and fix a bug preventing the generation of the NRPS/PKS products sidepanel.


kblin (approved):
Looks good to me 👍

PR #762, opened 2024-11-26
hmm_detection: add azoxy compound rules

SJShaw:
Adds azoxy compound rules for both dimerisation (e.g. azoxymycin) and carboxilic cross-link (e.g. azodyrecin) pathways.

I have sequences/alignments for the two custom profiles, which can be added to https://github.com/antismash/data-sources


kblin (approved):
Looks good to me 👍

PR #763, opened 2024-11-27
Refactor of common code and gene functions module

SJShaw:
Upcoming additional gene function tools need the gene function detection module to be shaped differently, so this PR sets that up to be ready to go. These changes also support the results having more detailed information than previous.

Some helper functions, e.g. diamond output parsing, were too specific to a module, so they've been converted to be more general. 

Some modules had useful functions that weren't in common, e.g. packing of HMM profiles into a single aggregated profile in `hmm_detection`, and they've been shifted into common.


kblin (approved):
Looks good to me 👍

PR #765, opened 2024-11-28
Misc small tweaks/fixes

SJShaw:
- Removes some dead code from lassopeptides
- Changes a number of general command line arguments to have an automatic inverse (e.g. `--no-debug` and `--no-minimal`) to handle cases where users would like to temporarily disable one of these options that they have enabled in a config file
- Fixes an issue in Clusterblast where self-hits weren't being consistently removed, leading to a crash in HTML generation


kblin (approved):
Looks good to me 👍

PR #768, opened 2024-11-29
Add an HTML helper to make clicking on a gene name highlight the relevant gene in the region viewer

SJShaw:
Requires antismash.js 0.16.1, up from 0.16.0.


kblin (approved):
Looks good to me 👍

PR #769, opened 2024-11-29
Extra tools for gene function classifications

SJShaw:
Adds a subset of [MITE](https://mite.bioinformatics.nl) data to classify tailoring enzymes using blast/diamond, with a minimum of 60% identity for a match.

Adds a tool that functions similarly to the existing SMCoGs, but with other profiles. This can be easily extended, unlike adding single profiles to SMCoGs. At some point, if a new SMCoGs set is built, any overlaps could be removed.


kblin (approved):
Looks good to me 👍

PR #770, opened 2024-11-29
Add terpene module

lisavader:
Add terpene analysis module


kblin:
There's dead code in the integration tests stopping me from merging this. Can you either use that where it should be used or remove it?

kblin:
there's also quite a number of additional pylint issues

antismash/modules/terpene/data/all_profiles.hmm
SJShaw, line 1: This should be in `.gitignore`, since it's already being rebuilt when necessary and just duplicates the other profiles.

antismash/modules/terpene/html_output.py
```diff
@@ -0,0 +1,99 @@
...
+def format_subtype(prediction: DomainPrediction) -> str:
+    if not prediction.subtypes:
+        if _hmm_properties[prediction.type].subtypes:
+            return "unknown"
+        return "none"
+    descriptions = []
+    for subtype in prediction.subtypes:
+        try:
+            short_description = "".join(_hmm_properties[subtype].description.split("; ")[1:])
+        except:
```
SJShaw, line 30: Bare `except` is dangerous, since it can also swallow syntax errors.

antismash/main.py
SJShaw, line 1: Don't change unrelated files in a single commit. Do it in a separate commit, if at all.

antismash/outputs/html/css/style.scss
```diff
@@ -1712,7 +1712,7 @@ ul.dropdown-options {
         * th, * td {
             padding: .8em 1em;
         }
-        thead tr v{
```
SJShaw, line 1715: Don't include unrelated changes in the same commit. This one in particular might have follow-on effects and it would be nice to know in the history when and why it was changed.

SJShaw (changes requested):
There's some unsafe pieces that need fixing, and a whole lot of linting and style issues.

Make sure you run `pylint` on the files you're changing, along with flake8`.

SJShaw (approved):

PR #771, opened 2024-11-30
tfbs_finder: add new matrices

HAugustijn:
Add new PWMs to the TFBS Finder based of data in the CollecTF database.

I constructed 33 new matrices, included a 'species' label in the details panel to indicate the species of origin for each PWM, and updated the JSON to include outgoing links to other databases.


SJShaw:
Now that there's so many names/entries that hit so often, I've added a commit to sort the glossary and truncate it to only strong and medium hits.

While I was there, I've also tweaked two of the things pylint complained about.

SJShaw (approved):

PR #776, opened 2024-12-06
Terpene module: Handle output for domains with same priority

lisavader:
This fixes cases where multiple domains with equal priority are being predicted.

- Changes the get_cluster_prediction function in terpene analysis:
When multiple domains of the same domain type have top priority, all of their sets of products are added to the cluster prediction instead of just one.
The order of the products is always the same.
Asserts that there are no two CompoundGroup instances with the same name.

- Adds a test for the get_cluster_prediction function


kblin (approved):
Looks good to me 👍

SJShaw (approved):

PR #777, opened 2024-12-09
terpene: Fix a template-related crash for regions with > 1 terpenes

kblin:
Reproducer for the crash is running `NC_003888.3` with `--taxon fungi`, which makes two of the terpene clusters get close enough to merge into a single region with the neighbourhood and cutoff boosts.


kblin:
@lisavader Can you please check that the template change makes sense for you?

SJShaw:
Superseded by #779

PR #778, opened 2024-12-11
Misc fixes

SJShaw:
Fixes a few bugs:
- the tailoring HTML tab included all tailoring hits in the whole record/contig, not just the relevant region
- cluster detection with `EXTENDERS` defined where the condition wasn't a subset of the rule's main conditions didn't mark any of those extra hits as core genes when within the area originally defined by the main conditions
- clusterblast crashed when generating javascript data for a region with gene identifiers being shared between query and reference (but not reference accession and query record identifier), largely due to some unused code that itself depended on an assert
- tfbs crashed for circular records when the intergenic cross-origin section had a match found only in the post-origin portion


kblin (approved):
Looks good to me 👍

PR #779, opened 2024-12-11
terpene: change HTML to be clearer with multiple clusters in a region

SJShaw:
Includes a variant of the fix in #777, but extends it to make the HTML output much clearer when multiple protoclusters are present.


kblin (approved):
Looks good to me 👍

PR #780, opened 2024-12-12
Some cleanup and new detection rules

kblin:


kblin:
Depending how we feel about adding more precursor rules, I'd have some more of those.

zreitz:
If they're highly specific to specialized metabolism, I'd probably be in favor of adding them.

SJShaw (approved):

SJShaw:
Merging what's here now, the extra precursor rules can follow as necessary.

PR #781, opened 2024-12-13
Add cytokinin detection rule

lisavader:
Adds a rule for detection of cytokinin BGCs.
This covers BGC0002163.
The gene cluster for fungal cytokinin biosynthesis was discovered in _Fusarium pseudograminearum_ and requires IPT (isopentenyl transferase) and LOG (lonely guy enzyme) domains. (https://doi.org/10.1111/mpp.12593)
Similar gene clusters have been found in other fungal genera (https://doi.org/10.1016/j.micres.2022.127103)
IPT and LOG domains are covered by PFAMs PF01715 and PF03641 respectively.
Based on a quick Swissprot search, the IPT domain is quite abundant in both bacteria and fungi, while there are only a handful of significant hits for the LOG domain.
As the LOG domain is the limiting factor for detection, it is not expected that the new rule will cause a lot of false positives.

This pull request contains:
- Add PFAMs PF01715 (IPT) and PF03641 (LOG)
- Add new strict rule 'cytokinin', which requires a hit to both IPT and LOG profiles


kblin:
Merged as part of #780

PR #782, opened 2024-12-13
HTML fixes

SJShaw:
Cross-origin genes were being truncated in the visualisation, showing only the pre-origin portion. This has been fixed here.

Page load times were in the order of seconds for some inputs. This turned out to be with RiPPs that had many CompaRiPPson results that generated quite a lot of HTML for the browser to parse.
A test input of six copies of nisin resulted in `index.html` being 2.3 MB, since each input had a precursor with around 130 matches to the antiSMASH-DB. Changing the rendering to only show the first 10 of the hits not shown by default reduced this file size to 383 KB and the load time to less than a second.

Semi-related: antiSMASH-JS has also been updated to 0.16.2 to fix an issue with multi-selecting, but that will apply automatically when running `download-antismash-databases`.

### DIFF
--- a/antismash/common/comparippson/templates/generic.html
+++ b/antismash/common/comparippson/templates/generic.html
@@ -38,6 +38,7 @@
 {%- endmacro -%}
 
 {%- set display_limit = 3 -%}
+{%- set extra_limit = 10 -%}
 <div class="comparippson-details">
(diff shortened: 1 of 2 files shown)


kblin (approved):
Looks good to me 👍

PR #783, opened 2024-12-13
Add sulfur transferase detection in gene functions

SJShaw:
Also fixes two issues, one with the 'extras' tool not annotating subfunctions, the other being the tailoring HTML tab only including subfunctions from SMCoGs.


kblin (approved):
Looks good to me 👍

PR #785, opened 2025-01-03
common.record_processing: catch missing sequences in fasta files via biopython

SJShaw:
Biopython allows FASTA files to have missing sequence information, e.g.
```
>R1
GCTA
>R2

>R3
TACG
```
This caused a crash in record processing, where the check for WGS records hit an `IndexError` in biopython's sequence indexing, that would previously have been caught in the following check for missing sequence information (which included WGS records with undefined sequences).

The fix here was to also handle that particular index error in the WGS check, leaving the original check for missing sequences in place as a backup for other potential code paths.


kblin (approved):
Looks good to me 👍

PR #788, opened 2025-01-15
Terpene minor updates

lisavader:
Various updates to terpene detection and analysis

Data-related:
* Hexaprenyl diphosphate (HPP) is not only a precursor to triterpenes, but also to the primary metabolite coenzyme Q and related quinones.
The existing profile PT_HPP selectively picks up quinone precursor biosynthesis genes, and as this is primary metabolism (not terpene related), this profile was excluded in the terpene rule and dropped from terpene analysis.
Instead, a new loose rule was made called 'quinone_isoprenoid_chain' to catch these genes.
* The HAD_2 profile is not specific enough to predict hydroxylation of C2-C7/C6-C11 sesquiterpenes, therefore HAD_2 was dropped from the analysis module (but is still a related profile of the terpene rule)
* Products of the T1TS_KS profile were updated to be more specific
* The C6-C11 sesquiterpene product was removed from T2TS_C15_b, because this product type is only reported in the T2TS_C15_a group and not in the T2TS_C15_b group
* Adjustments to compound and biosynthetic subclass names

Html-related:
* Added the functional group abbreviation 'PP' to the glossary

Other:
* Fixed a mistake in the terpene test file

In this pr, .hmm files within the terpene module are converted to hmmer2 file format to support later updates requiring this file format


lisavader:
I added a very minor html update; when the terpene module finds no domains at all this is now explicitly reported.

antismash/detection/hmm_detection/cluster_rules/loose.txt
lisavader, line 1: Hi Kai,
I remember you once wrote down an updated rule name and description for this.

SJShaw (approved):

PR #790, opened 2025-01-21
Misc fixes

SJShaw:
Fixes a few issues found in testing and use:

- a crash when a circular plasmid had a cross-origin protocluster core that didn't cover the full plasmid, until the neighbourhood was added
- fixes protein to DNA coordinate conversion from parent features with ambiguous positions (e.g. `<10..50`) creating locations with both offset positions being the same ambiguous type (e.g. `<15..<45`)
- renames the `azdH_alignment` profile, removing the unnecessary `_alignment` portion of the name


kblin (approved):
LGTM 👍

PR #791, opened 2025-01-21
terpene: Add specific profiles for Pyr4 type terpene cyclases

lisavader:
I added profiles for specific types of fungal Pyr4 type terpene cyclases:
- Polyketide sesquiterpenoid synthases
- Polyketide diterpenoid synthases
- Indole diterpenoid synthases
- Onoceroid (= triterpenoid) synthases

The detection of Pyr4 type terpene cyclases remains the same, but the terpene analysis module now contains 6 extra profiles covering the categories above.
The sequences used to construct these profiles were retrieved from the database funBGCs (https://personal.cityu.edu.hk/ymatsuda/funbgcs/funbgcs.html)
The annotation of these sequences is from [Tang & Matsuda (2024)](https://www.nature.com/articles/s41467-024-48771-7)


kblin:
This seems to overlap with #788

lisavader:
Yes, that's right, I rebased this branch on top of #788.
In #788 I changed the format of the hmm files from hmmer3 to hmmer2, this is for compatibility with a coming update that uses hmmpfam.
Hmmpress doesn't take files in different file formats, so therefore the Pyr4 profiles also need to be in hmmer2 format.

SJShaw (approved):

PR #792, opened 2025-01-22
hmm_detection: add leupeptin rule

SJShaw:
Rule and profiles for detecting leupeptin [BGC0002134](https://mibig.secondarymetabolites.org/go/BGC0002134).


kblin (approved):
Looks good to me 👍

PR #800, opened 2025-03-02
Fix: parse circular topology from gff

### LINKED ISSUE #713: Error: could not parse records from GFF3 file
Hi there,

I am running antismash v7.1.0 downloaded via bioconda. I have installed the antismash database using the download-antismash-databases command. When I run the antismash command, I get the following error within 10 seconds whether I provide a gff3 file or not:

```ERROR: could not parse records from GFF3 file```

**To Reproduce**
Here are the two commands that each produce this same error:
```antismash --databases /staging/groups/gluckthaler_group/databases/antismash --cpus 4 --output-basename test --output-dir output/ --taxon fungi --clusterhmmer --verbose --genefinding-tool glimmerhmm fusoxy65.fna```

```antismash --databases /staging/groups/gluckthaler_group/databases/antismash --cpus 4 --output-basename test --output-dir output/ --taxon fungi --clusterhmmer --verbose --genefinding-gff3 fusoxy65.gff3 fusoxy65.fna```

**Screenshots**
contents of database directory: database_content
contents of gff3 and assembly file:
file_content
GFF error 1:
gff_error1
GFF error 2:
gff_error2

**System (please complete the following information):**
CentOS Stream 9

**Additional context**
I also tried installing the databases in their default location within the conda package directory, but I get the same error

What could be going on here? Thanks for your help

### LINKED ISSUE #796: could not parse records from GFF3 file, antiSMASH 7.1.0
**Describe the bug**

I am running antismash on a FASTA file (without genome annotation), and some of the contigs in the file are short.

I suspect antismash tries to run gene prediction (in my case **glimmerhmm** since I use `--genefinding-tool glimmerhmm --taxon fungi`) and since it can't find any genes on a specific contig sequence, it quits with the error: 

`ERROR: could not parse records from GFF3 file`

The problem is that probably the whole antismash run and gene search process is terminated because of missing genes on a contig -> no GFF file is generated -> `ERROR: could not parse records from GFF3 file` (because there is no GFF file?).

Above just my guess. The output folder of antismash is empty.

**Expected behavior**
Continue searching for genes on other contigs and do not terminate even if some contigs have no detected genes.

Makes sense? Apologies if I am missing something here.

Best,
Gregor

antiSMASH 7.1.0

### LINKED ISSUE #799: Bug Report: USELESS and PUZZLING check when using NCBI reference GFF
When annotating complete genome such as `GCA_025309455.1`, antismash always fail like this:

```log
ERROR    01/03 23:42:38   GFF3 record and sequence coordinates are not compatible.
ERROR    01/03 23:42:38   incompatible GFF record and sequence coordinates
ERROR: incompatible GFF record and sequence coordinates
```

It was raised here:
https://github.com/antismash/antismash/blob/e448aadc9bebc562cdf4f615baafcbe68a312c6c/antismash/common/gff_parser.py#L63-L65

However, while the genome is only 5111300 bp long, the last CDS is just range from 5109461 to 5115166 as it is a complete genome. 
```log
CP104377.1      Genbank region  1       5111300 .       +       .       ID=CP104377.1:1..5111300;Dbxref=taxon:2977320;Is_circular=true;Name=ANONYMOUS;collection-date=2021-10-05;country=China: Nanjing City%2C Jiangsu Province;gbkey=Src;genome=chromosome;isolation-source=gut;mol_type=genomic DNA;nat-host=Procambarus clarkii;strain=PR12
...
CP104377.1      Genbank gene    5109461 5115166 .       +       .       ID=gene-N4T19_00005;Name=N4T19_00005;gbkey=Gene;gene_biotype=protein_coding;locus_tag=N4T19_00005
CP104377.1      GeneMarkS-2+    CDS     5109461 5115166 .       +       0       ID=cds-UXC18563.1;Parent=gene-N4T19_00005;Dbxref=NCBI_GP:UXC18563.1;Name=UXC18563.1;gbkey=CDS;inference=COORDINATES: ab initio prediction:GeneMarkS-2+;locus_tag=N4T19_00005;product=phage tail length tape measure family protein;protein_id=UXC18563.1;transl_table=11
```

This check is just nonsense.

Hocnonsense:
This is a fix to accept cross-end features on circular genomes (fix #799)

This fix also allow a "force correlation" case when there are only one genome and one gff file, and parse features to record in the case.
~A more strict check is applied on each pair of sequence and gff, not only for the "force correlation" case.~
~The GFF reader library is updated for better performance and newer Biopython support.~


SJShaw (changes requested):
There's a number of things I'd like to see changes for before even getting into a real code review:

- your autoformatter introduces an awful lot of irrelevant changes, e.g. whitespace-only changes to lines
- you're missing tests to cover these changes
- you're missing docstrings for the functions you're adding
- you're ignoring the style of the existing code and creating your own (e.g. choosing to make module level constants lower case)
- you've added a random print statement, presumably left over from debugging

Please, fix those up to make it easier to do a real code review.

Hocnonsense:
Sure. I'll clean these code and add tests&docs as quick as possible. Sorry to bother you.

However, may I ask which formatter is used for antismash? my default formatter black seems to be not suited here.

SJShaw:
We don't have a particular autoformatter. While I'm sure one could be configured to match most of what we have, we have some older styles in some places for which don't want to automatically alter line history.

You can use one if you like, but then use `git add -p` and only add the changes relevant to the commit.

Hocnonsense:
Sure. here are the updates:

- [x] revert changes introduced from autoformatter, following the existing code.
    - [x] change the private module level constants `_BIOPYTHON_STRAND` to upper case
    - [x] new line in brackets will start after the same column of the left braket, or just follow them as is
    - [x] docstring format
    - [ ] The only thing I persist in making modifications is the random use of single bracket and double bracket. Hope this change can make the file eazier to read in the future.
- [x] add test to cover changes: `GffParserTest::test_circular` to check dealing of gff file with circulre featuer
- [x] add docstrings for the new function `check_gff_suitability`
- [x] clean useless `print` left over from debugging

I hope the quality of this pull request is up to standard.

SJShaw:
I don't know if this is you doing this manually or you're still using some kind of tool that's making this automatic, but you're swapping a bunch of single quotes to double quotes _and_ swapping double quotes to single quotes in other places.

Look at the "Files changed" tab of the PR, then go through then and remove _all_ unrelated changes.

antismash/common/gff_parser.py
```diff
@@ -21,6 +22,40 @@
...
+    qualifiers.update(feature.attributes)
+    start, stop = sorted((feature.start, feature.end))
+    return SeqFeature(
+        SimpleLocation(start - 1, stop, strand=_BIOPYTHON_STRAND[feature.strand]),
+        id=feature.id,
+        type=feature.featuretype,
+        qualifiers=qualifiers,
+    )
+
+
```
SJShaw, line 58: This is almost an exact copy/paste of `gffutils.biopython_integration.to_seqfeature`. You've removed the string handling portion and have added type annotations, but have left the docstring saying that a string is accepted. I don't know why you have done this instead of just using the existing function from the library you've just added as a dependency.

This is a licensing minefield. Before doing anything else with this PR, we will need to know what else you've copied from other repositories.

kblin:
So without having looked at the code yet, I remember looking at `gffutils` some years ago and then deciding to stick with `bcbio-gff` instead because (at the time) it worked better for our use cases. I can't remember off-hand what my gripe with `gffutils` was, but I'm a bit skeptical of changing out parsing libraries until I remember.

It also seems unrelated to fixing the GFF topology issue, so why don't we focus on that first?

antismash/common/gff_parser.py
Hocnonsense: Sure, there is almost the same between `gffutils.biopython_integration.to_seqfeature` and this function, but in `gffutils`, it checked `FeatureLocation`, which was once deleted in biopython and marked as `OBSOLETE` now.

https://github.com/daler/gffutils/blob/67a4cd7edc11bb2d1ff97d8c777fb1e6006296e5/gffutils/biopython_integration.py#L6C1-L11C68

Hocnonsense:
> It also seems unrelated to fixing the GFF topology issue, so why don't we focus on that first?

I proposed to using `gffutils` instead of `bcbio-gff` for several reasons:
1. faster, especially when you input a large gff with many fasta records in it
2. more modern
3. I noticed that the dependence is `bcbio-gff==0.7.1` in current version, however, it is incomparable to newer Biopython (#713 and #796)

antismash/common/gff_parser.py
SJShaw: That doesn't change potential licensing issues. 

> This is a licensing minefield. Before doing anything else with this PR, we will need to know what else you've copied from other repositories.

Please, address the above.

Hocnonsense: Sure. I'm sorry for this licensing minefield.

I'm sure that none of the other codes were copied or modified from elsewhere. For test data, I modified the example of bacteriophage f1 J02448.

Hocnonsense:
> So without having looked at the code yet, I remember looking at `gffutils` some years ago and then deciding to stick with `bcbio-gff` instead because (at the time) it worked better for our use cases. I can't remember off-hand what my gripe with `gffutils` was, but I'm a bit skeptical of changing out parsing libraries until I remember.
> 
> It also seems unrelated to fixing the GFF topology issue, so why don't we focus on that first?

Thanks for your comments. Sorry for trying touch multipile topics in the single issue. 
I've revert such modifications and focus on topology only. I think I should push that one here.

SJShaw (changes requested):
Normally, I'd test out significant input handling changes on a wide variety of the typical inputs that antiSMASH users tend to provide.

However, as is, this pull request is dangerous for antiSMASH to include, so I haven't yet tried that.

It might be faster for me to make a separate PR  to support this instead of repeatedly requesting changes which then aren't fully tested.

PR #801, opened 2025-03-03
Fix a crash when cluster neighbourhoods wrap more than once around the record

SJShaw:
With a very small circular record/contig, e.g. 38kb long, a protocluster with a cross-origin core could have a neighbourhood (e.g. 20kb each side) applied such that the neighbourhood extension alone was longer than the record. 
This caused the location extension to always a single part full-record location, which then crashed in the check that any protocluster with a cross-origin core must also have a cross-origin neighbourhood.

To solve this, an option to the relevant extension function was added to force the extended location to also be cross-origin, taking the mid-point between the existing location's start and end points as the point of separation.

In order to not appear as a point location (i.e. start and end are equal), the end coordinate is shifted back by one.
This in turn caused the containment checks of protoclusters against their defined `SUPERIORS` to return false when they did _not_ cross the origin and the superior _did_ cross the origin with this full-record neighbourhood. The solution there was to change the cluster containment check to compare not the location which included the neighbourhood, but only the core.

### DIFF
--- a/antismash/common/hmm_rule_parser/cluster_prediction.py
+++ b/antismash/common/hmm_rule_parser/cluster_prediction.py
@@ -8,6 +8,7 @@ import bisect
 from collections import defaultdict
 import dataclasses
 import logging
+import math
 from typing import Any, Dict, FrozenSet, Iterable, Iterator, List, Optional, Set, Tuple, Union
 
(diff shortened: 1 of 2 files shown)


kblin (approved):
Looks good to me 👍

PR #803, opened 2025-03-11
common.gff_parser: handle circular annotations in landmark features

SJShaw:
A replacement PR for #800 to handle circular annotations in GFF files, fixing #799.

- updates a record's topology when a GFF landmark feature marks it as such
- splits cross-origin feature locations into parts where necessary
- aborts if a record's topology is already explicitly defined (e.g. from an unannotated genbank file) and the GFF file explicitly sets a different topology
- adds the landmark feature as a `source` feature within `secmet.Record`, which previously wasn't added


kblin (approved):
Looks good to me 👍

PR #804, opened 2025-03-13
New hmm hit filtering methods

lisavader:
* New method for fragment merging
  A single gene might contain multiple standalone domains of the same type (e.g. geosmin synthase, consisting of two type 1 terpene synthase domains).
  The current hmm hit merging function doesn't support this, because hits to the same profile are always merged (provided this doesn't result in a hit larger than 1.5x the profile length).
  Therefore, a new 'fragment mode' was added to `merge_domain_list()`, which merges fragments of a single domain, but not standalone domains.

* New mode for filtering hmmscan results
  The terpene analysis module requires a specific filtering method for hmm hits, where merging is done in fragment mode and overlapping hits are preserved.
  This new 'preservation mode' was added to `refine_hmmscan_results()`.

* Changes to properties and methods of HMMResult
  To support these changes, the properties `hit_start` and `hit_end` were added to the HMMResult class.
  `__len__()` has now become ambiguous and is replaced by the `query_length` and `hit_length` properties.
  Similarly, the methods `overlaps__with()` and `is_contained_by()` have become ambiguous and are replaced by specific methods for the query region and the hit region.
  It is now possible to specify a minimum overlap value in the overlap methods.

### DIFF
--- a/antismash/modules/terpene/terpene_analysis.py
+++ b/antismash/modules/terpene/terpene_analysis.py
@@ -9,7 +9,7 @@ from typing import Iterable
 from collections import defaultdict
 
 from antismash.common import fasta, path
-from antismash.common.hmmscan_refinement import HMMResult, QueryResult, gather_by_query, remove_incomplete
+from antismash.common.hmmscan_refinement import HMMResult, QueryResult, refine_hmmscan_results
 from antismash.common.secmet import Protocluster, CDSFeature
 from antismash.common.subprocessing.hmmscan import run_hmmscan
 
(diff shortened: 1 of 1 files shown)


lisavader:
Tilmann suggested that the fragment merging mode might also be useful for pks clusters, because it is possible to find multiple domains of the same type right next to each other, and you don't want to merge them in this case.
This is an example he gave with CPs occuring next to each other (although they are not being merged in this case, I guess because they are too far apart?):
https://antismash.secondarymetabolites.org/upload/bacteria-c2d8624f-e087-4cfe-bb1c-3a6580617617/index.html#r1c1
What do you think about this @kblin?

PR #806, opened 2025-03-17
genefunctions: show explicit not-found message in HTML rather than empty

SJShaw:
Empty output sections are ambiguous as to whether there was a fault or no results. This adds an explicit message to clarify it.

Before:
image

After:
image


kblin (approved):
Looks good to me 👍

PR #807, opened 2025-03-27
hmm_detection: add relevant RRE profile as extender for lassopeptides

SJShaw:
There's [many](https://antismash-db.secondarymetabolites.org/area.html?record=NC_006322.1&start=3863358&end=3863484) [examples](https://antismash-db.secondarymetabolites.org/area?record=NZ_AUJV01000005&start=40770&end=75703) of `lassopeptide` protoclusters being beside `RRE-containing`, where the RRE hit is to `Stand_Alone_Lasso_RRE`.

It seems simpler to add that RRE profile to the more defined rule in the `EXTENDERS` section, so those "hybrids" will form a single contiguous protocluster instead.

Examples of affected hybrids that now form a single protocluster:
image
image


kblin (approved):
Looks good to me 👍

PR #808, opened 2025-03-27
Fixes for non-standard and incorrect HTML formatting

SJShaw:
There's a number of issues that weren't legal HTML, even though they were rendering correctly (so far). Among them are:

- elements without closing tags
- closing tags without opening tags
- mismatched opening and closing tags
- incompatible child elements (e.g. `div` in a `span`)
- etc

Also changes "Confidence" in the HTML overview table to "Similarity Confidence" due to ambiguity around it's application to similarity or the protocluster calls themselves.


kblin (changes requested):
One minor change to the one alt text, looks fine otherwise. :+1:

kblin (approved):
Looks good to me 👍

PR #809, opened 2025-03-27
config: add flag to skip gbk summary writing

zreitz:
The genbank summary file is the single largest file in the output directory, adding up to gigabytes across my high-throughput runs, and none of my downstream tools use it.

I added a flag `--skip-gbk-file` to stop its creation. Modeling it after `--skip-zip-file`, I added two `if` statements to stop creation and to remove the download option from the HTML.

### DIFF
--- a/antismash/config/args.py
+++ b/antismash/config/args.py
@@ -668,11 +668,11 @@ def debug_options() -> ModuleArgs:
                      action=argparse.BooleanOptionalAction,
                      default=False,
                      help="Skip input record sanitisation. Use with care.")
-    group.add_option('--skip-zip-file',
-                     dest='skip_zip_file',
+    group.add_option('--zip-output',
+                     dest='zip_output',
                      action=argparse.BooleanOptionalAction,
-                     default=False,
-                     help="Do not create a zip of the output")
+                     default=True,
+                     help="Create a ZIP file of the output (default: %(default)s)")
     group.add_option('--summary-gbk',
                      dest='summary_gbk',
                      action=argparse.BooleanOptionalAction,
--- a/antismash/config/args.py
+++ b/antismash/config/args.py
@@ -678,6 +678,11 @@ def debug_options() -> ModuleArgs:
                      action=argparse.BooleanOptionalAction,
                      default=True,
                      help="Create a GenBank summary file (default: %(default)s)")
+    group.add_option('--region-gbks',
+                     dest='region_gbks',
+                     action=argparse.BooleanOptionalAction,
+                     default=True,
+                     help="Create a GenBank file for each region (default: %(default)s)")
     return group
 
 
--- a/antismash/config/args.py
+++ b/antismash/config/args.py
@@ -673,6 +673,11 @@ def debug_options() -> ModuleArgs:
                      action=argparse.BooleanOptionalAction,
                      default=False,
                      help="Do not create a zip of the output")
+    group.add_option('--summary-gbk',
+                     dest='summary_gbk',
+                     action=argparse.BooleanOptionalAction,
+                     default=True,
+                     help="Create a GenBank summary file (default: %(default)s)")
     return group
 
 
--- a/antismash/main.py
+++ b/antismash/main.py
@@ -459,10 +459,11 @@ def write_outputs(results: serialiser.AntismashResults, options: ConfigType) -> 
     # add antismash meta-annotation to records
     add_antismash_comments(list(zip(results.records, bio_records)), options)
 
-    logging.debug("Writing cluster-specific genbank files")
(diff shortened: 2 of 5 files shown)


SJShaw:
I think this is a perfectly reasonable feature to add.

I would probably changing the naming and flip the default value of the argument though. `--no-skip-gbk-file` is a bit clunky compared to `--no-gbk-output`. Swapping `file` to `output` here is only because I would expect to specify a file after an arg ending in "file".

zreitz:
Yeah, I'm definitely open to a name change.

It's still producing the cluster gbks (they're small and useful for e.g. BiG-SCAPE), so `--gbk-outputs` isn't accurate. Perhaps `--gbk-summary` / `--no-gbk-summary`  for clarity?

Side note, would you approve similarly renaming `--skip-zip-file` to `--zip-output` / `--no-zip-output` in a separate PR or do you want to avoid changing an existing flag?

lisavader:
Hi Zach,
This is great, I was actually just having the same issue with redundant output files being generated!
I think it would make sense to also add an option for skipping to region .gbk files creation (something like `--no-region-gbks`).
In the end, all output files can be generated from the .json file, so from a storage perspective it makes sense to just store the json.
Maybe we could even add a shorthand `--json-only` that does the same as three flags `--no-zip-output`, `--no-gbk-summary` and `--no-region-gbks` (or equivalent names).

SJShaw:
After thinking a little more about this, there's going to be ambiguity/a precedence issue with having a `--json-only` _and_ any of the other general output options. 

For example: if a user has `--json-only` in their config file, then decides to get some region gbks, they may well be confused why add `--region-gbks` doesn't generate any. Conversely, if there's an explicit `--no-skip-size` in the config file and then `--json-only` is added on invocation, the zip file will still be created.

This is similar to `--minimal`, where a extra handling is required to modify options for those modules which are enabled by default. The problem is that there's no module args concept for antiSMASH-wide options that allows for that uniform approach.

zreitz:
>Conversely, if there's an explicit `--no-skip-size` in the config file and then `--json-only` is added on invocation, the zip file will still be created.

I don't know what " `--no-skip-size` in the config file" means.

`--json-only` should prevent zip creation as it totally stops `write_outputs()`.

>Side note, would you approve similarly renaming --skip-zip-file to --zip-output / --no-zip-output in a separate PR or do you want to avoid changing an existing flag?

At this point, I think this change should go in this PR to put all of the flags in line.

SJShaw:
> > Conversely, if there's an explicit `--no-skip-size` in the config file and then `--json-only` is added on invocation, the zip file will still be created.
> 
> I don't know what " `--no-skip-size` in the config file" means.
> 
> `--json-only` should prevent zip creation as it totally stops `write_outputs()`.

Whoops, I meant `no-skip-zip` there. It cou

> > Side note, would you approve similarly renaming --skip-zip-file to --zip-output / --no-zip-output in a separate PR or do you want to avoid changing an existing flag?
> 
> At this point, I think this change should go in this PR to put all of the flags in line.

I agree. I think it's a major version change to make and making them consistent is good.



> I was actually going to bring up `--minimal` later, after discovering that `--no-enable-html` cannot possibly do anything. Could you explain the motivation behind adding explicit disablings with `BooleanOptionalAction`? In my mind, the clearest solution is to revert back to `store_true` when there are these issues.

The toggle option is specifically in place to handle config files enabling an option that's false by default and then trying to re-disable it via the command line. For example, I have `--debug` set in my config file because I want to always see it while doing dev work, but then I don't want to have to change my config file every time I want to check out what logging looks like with just `--verbose` or no logging at all.

`--minimal` was brought in to have a similar function as 4's "fast mode", which skipped some (very) long analyses that were on by default. It's morphed slightly since it was introduced, since we didn't originally have the option to disable it at all. It's now a little more ambiguous, but it allows for explicit overrides to re-enable default modules as well as enabling optional modules.

I'd like to change it all so that there's `--cb-known` and the opposite `--no-cb-known` for the same config file issues, but it just hasn't happened yet.

zreitz:
Thanks for the explanation, that all makes sense. 

For now I've removed the `--json-only` flag (sorry Lisa). IMO it's not worth the trouble to just write three fewer flags.

zreitz:
This does mean that there's no way to disable html output without invoking `--minimal`, but I think we accept that for now

SJShaw:
I've just rebased the PR to remove the reverted commits entirely and squash those that are paired.

I also fixed one of the changed lines that had a tiny existing issue while it was there, a link URL missing the wrapping quotes. All you'd changed was the indent, which is absolutely fine.

SJShaw (approved):

PR #815, opened 2025-04-08
Rare crash and rejection fixes

SJShaw:
Fixes a crash when user inputs had sequences longer than 32 megabases and required/requested genefinding via prodigal. Prodigal's maximum sequence length is 32 megabases. Sequences such as these are now rejected and users can instead submit sequences that have already had genes annotated. As part of testing this, the `run_prodigal` module was renamed to `prodigal` to avoid conflicts with module and function names.

Fixes a crash in clusterblast HTML generation when handling circular records with a full record region and a cross-origin CDS.

Changed the CDS translation validity test to be more relaxed, allowing for translations longer than the given coordinates support when the start is ambiguous. This matches the existing behaviour for allowing translations to be longer when the _end_ coordinate was ambiguous.


kblin (approved):
Looks good to me 👍

PR #819, opened 2025-04-23
Extend the detection of Pyr4 type terpene cyclases

lisavader:
I noticed that we can lower the cutoff of the TS_Pyr4 profile to enable detection of bacterial homologs (see [doi.org/10.1002/anie.202422270](url), GroF), without noticeably increasing the false positive rate.
I scanned TS_Pyr4 against all proteins in the antismash DB, and it hits some transporters with bitscores <20.
Other than that there is a steep jump to hypothetical proteins with bitscores >70, which I expect to be actual terpene cyclases.


kblin (approved):
Looks good to me 👍

PR #820, opened 2025-04-24
secmet: fix false negatives in cross-origin test in linear records

SJShaw:
Fixes some crashes resulting from whole-contig genes that contained exons. The tests for cross-origin genes in linear records made some faulty assumptions about start and end coordinates with multiple exons.


kblin (approved):
Looks good to me 👍

PR #821, opened 2025-04-24
clusterblast: fix outputs using incorrect locations of reference genes with duplicate identifiers

SJShaw:
This issue cropped up most with knownclusterblast, due to a number of MIBiG entries having used prodigal for genefinding, which resulted in many copies of names like `ctg1_orf1`. The lack of database-wide identifiers caused both summary and SVGs to use the location of the latest reference gene in the database with that name. This made some reference clusters look very unlike the query, despite potentially being a perfect match.

The scoring and ranking is unchanged.


kblin (approved):
Nice trick with the test datadir, took me a moment to understand but is much cleaner.

PR #823, opened 2025-04-29
Fix for region features in region genbanks not referencing correct candidate numbers

SJShaw:
Candidate cluster numbers in region genbanks are adjusted to start counting at 1. This was correct for the candidate cluster features themselves, but not in the references within the region feature's qualifiers.


kblin (dismissed):
Looks good to me 👍

kblin (approved):
Still looks good to me 👍

PR #825, opened 2025-04-29
Misc python 3.12/macos fixes

SJShaw:
- 3.12 changed regex boolean evaluation and one of the options (`len(node)`) isn't supported in 3.11
- handles newer macOS system calls for python parallelism failing to serialise our config
- fix test using `cat` that expected the same failure message on both linux and mac


kblin (approved):
Looks good to me 👍

PR #826, opened 2025-05-02
Fixes for strange cases in input handling

SJShaw:
One fix is trivial, changing an error type so that a normal input error is raised instead of ending up with a stack trace.

The other fix is for an extremely specific style of input, requiring:
- a FASTA input
- duplicate first "word" in identifiers (these split on whitespace); with
- one of the duplicates has a trailing character considered invalid for record identifiers (e.g. `!` or `,`)
- `--skip-sanitisation` enabled (the default)

The root cause was that stripping the invalid characters occurred after the deduplication of naming, which then override any results keyed by identifier and caused a crash later in a few places where the number of regions/clusters was different to the record with the name collision.


kblin (approved):
Looks good to me 👍

PR #832, opened 2025-05-08
config: add an option to remove all existing annotations from an input

SJShaw:
I regularly come across Genbank/RefSeq entries that have poor gene calling with PGAP and PKS genes. I'd normally have to convert the GBK to FASTA, then run antiSMASH over the result.

This PR brings in `--remove-existing-annotations`, which removes every feature from the inputs, allowing other genefinding to be used. The inverse is also accessible, `--no-remove-existing-annotations`, which isn't great naming but I don't have better ideas on that.


kblin (approved):
Looks good to me 👍

PR #836, opened 2025-05-15
Small corrections for input handling

SJShaw:
- wraps plain `ValueError` stack traces for GFF feature annotations outside the matching FASTA record
- prevents coercing translations from starting with `M` in cases where the feature's start location is ambiguous


kblin (approved):
Looks good to me 👍

PR #838, opened 2025-05-20
Fixes for output and data generation

SJShaw:
Writing outputs could fail when using `shutil.copytree` where source directories had no owner write permissions and then additional files were added at a later point (e.g. adding `antismash.js` to the templated `js` subdirectory in the output). This also affected any integration test that used `antismash.common.test.helpers.run_and_regenerate_results_for_module()`, where the temporary directory used as an output would then not be able to clean up after itself.


kblin (approved):
Looks good to me 👍

PR #842, opened 2025-06-18
Fixes for odd/unexpected input related errors

SJShaw:
Diamond raises an error when a protein sequence looks like a nucleotide sequence. This happened with some tiny pseudogenes, where the translation was valid. Diamond can be set to ignore warnings, and since this is the only case covered by that flag (as of now), it's been added.

When an input was circular and the sequence around the origin contained a small, unannotated, gene that were then found in that area with `all_orfs` were then incorrect when the gene was trimmed to the shortest length (i.e. latest start codon). The trimming function didn't consider these cross-origin genes and could result in a gene covering the entire contig, and in some cases crashing due to the translation of that gene being longer than the allowed maximum.

Some inputs crashed inside cluster prediction when a cluster was detected for a cluster with `EXTENDERS` _and_ they had a particular combination of genes at the origin.

Fungal sequences were assumed to never be circular, but some NCBI assemblies include mitochrondions which are circular. This resulted in incorrect detection of exons being in inconsistent ordering and raising an error. This is made more difficult since most fungal inputs are FASTA + GFF, where circularity isn't always specified, but for the moment all inputs are now considered as possibly circular.

Inputs with a cross-origin gene that had `codon_start` qualifiers would crash.

Due to an off-by-one error, GFF parsing would crash when a cross-origin gene had an exon beginning precisely at the origin.

Lastly, but most complicated, is a crash/error when a feature had overlapping exons/sections. This was intended behaviour, but some NCBI assemblies do contain features that have this case. That requirement has been removed, which in turn led to the fetching of contig coordinates from protein coordinates within a feature being wildly incorrect. This accounts for most of the line changes.

### DIFF
--- a/antismash/common/all_orfs.py
+++ b/antismash/common/all_orfs.py
@@ -70,14 +70,23 @@ def get_trimmed_orf(orf: CDSFeature, record: Record, include: int = None,
         return None
 
     # otherwise, use the start which gives the smallest possible ORF
+    start = orf.start
+    end = orf.end
     if orf.location.strand == 1:
-        start = orf.location.start + starts[-1]
-        end = orf.location.end
+        start += starts[-1]
+        start %= len(record)
     else:
-        start = orf.location.start
-        end = orf.location.end - starts[-1]
+        end -= starts[-1]
 
-    location = FeatureLocation(start, end, orf.location.strand)
+    if start > end:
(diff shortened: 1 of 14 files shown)


SJShaw:
Whoops, that reference to this issue in a new PR was me getting the issue number wrong. Ignore it.

PR #843, opened 2025-06-19
secmet: fix protocluster sorting non-determinism when locations are equal

SJShaw:
`Record.from_biopython()` and regeneration of a `Record` from existing JSON results could cause ordering differences when locations were the same. This changes the sort order to include the product name in that situation to make it (more) deterministic.

This can't cover cases where just a location is used as a comparator, but `Record.add_protocluster()` always compares protoclusters with full instances and not just locations.


kblin (approved):
Looks good to me 👍

PR #844, opened 2025-06-26
Split 'minimum' keyword in rules to handle with repeats and without

SJShaw:
`minimum(2, [a, b,c])` is intuitively "a minimum of 2 of a, b, or c".
The actual interpretation is "a minimum of 2 _CDSes_ with a, b, or c".

Some rules have used it with one interpretation and some have used it with the other.

To handle this, the more intuitive naming has been changed to the stricter form where the count is of distinct hits within the group. The existing form, where the number of distinct hits isn't relevant, has been reintroduced as `repeatable`. The decision to rename the existing functionality instead of the stricter version is that the intuitive interpretation will continue to confuse people and not be obvious when testing new rules.

The saccharide rules, `saccharide` and `oligosaccharide`, are now explicitly repeatable. No cluster previously detected will be lost.

The only other rule affected is `isocyanide-nrp`, which keeps the `minimum` keyword and now requires distinct hits. This will result in some previously detected clusters no longer being detected, however in this case it is not expected that multiple CDS features with _only_ `PP-binding`  domains would cause a cluster to be called.

### DIFF
--- a/antismash/common/hmm_rule_parser/rule_parser.py
+++ b/antismash/common/hmm_rule_parser/rule_parser.py
@@ -88,6 +88,7 @@ The grammar itself:
     ID = [a-zA-Z]{[a-zA-Z0-9_-]}*
     TEXT = {[a-zA-Z0-9_-.]}*
     MINIMUM_LABEL = 'minimum'
+    REPEATABLE_LABEL = 'repeatable'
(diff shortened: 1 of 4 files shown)


kblin (approved):
Looks good to me 👍

PR #846, opened 2025-07-30
hmm_detection: add rule for aureonuclemycin

drboothtj:
Created a new rule for **herbicidin** detection (specifically, the **aureonuclemycin** scaffold, see: https://pubs.acs.org/doi/10.1021/jacs.2c05728).

To create the profiles, I identified conserved clusters of the 4 core genes (Her4-7) in the NCBI database using cblaster. Cases where hits were duplicated were omitted and the remaining single copy orthologous were used to make the models. 

The bitscores were estimated based on the distribution of hit using the online HMMsearch versus the reference proteome database (https://www.ebi.ac.uk/Tools/hmmer/search/hmmsearch). These bitscores are relatively high compared to other rules, but identified herbicidins and several variants.

CUTOFF and NEIGHBOURHOOD values were arbitrarily set to 10 kb.


kblin (approved):
Looks good to me 👍

PR #847, opened 2025-08-01
Adding more support for use of antiSMASH as a library

SJShaw:
**NOTE**: this is very much a work in progress, in that there's quite a lot of finer details still not covered, but should serve as a proof of concept.

There's a few variants of antiSMASH out there at this point. Most of them are forks that are then edited as needed, which leads to quite a lot of issues when attempting to update with bug fixes and new features coming into antiSMASH itself.

This is a first pass at modifying much of antiSMASH to allow for use as a library without nearly as much monkey patching.
For example, before these changes, adding custom modules for a variant using antiSMASH's full pipeline required, as a minimum:
```python
import antismash
antismash.main._ANALYSIS_MODULES = [custom_module, ...]
antismash.outputs.html = custom_html

antismash.__main__.main(cmdline args)
# or correctly building a config object, then
# antismash.main.run_antismash(input_file, options)
```
and then quite a lot of find/replace in output files.

For that same use case and with these changes, that would instead be:
```python
import antismash
antismash.main.replace_analysis_modules([custom_module, ...])
antismash.main.replace_output_modules([custom_html, ...])
antismash.__main__.main(cmdline args, branding_override="customSMASH", version_override="1.0")
```
That would then run the given modules and have all previous mentions of `antiSMASH 8.something` instead reported as:
`customSMASH 1.0 (based on antiSMASH 8.something)`

It's expected that any existing modules can be reused and that any new modules would be able to import any current code and reuse it without issue (e.g. the clusterblast analyses are currently difficult to use with custom databases and the code itself has a number of hardcoded limits that may need changing.

There's also a lot more involved to avoid more monkey patching, including:
- changing default user options files (e.g. `antismash8.cfg` always being used without more monkey patching)
- inability to reuse config loading when `instance.cfg` files were used
- broken library usage everywhere code referred to `__file__`
- the HTML module assuming some other modules existed
- hardcoded/magic numbers need to be accessible somehow
and so on.

There's at least one real project using this as a base, but it needs to be better formalised, have more complete testing, and obviously complete the remaining portions which still aren't finished.

### DIFF
--- a/antismash/modules/clusterblast/core.py
+++ b/antismash/modules/clusterblast/core.py
@@ -141,28 +141,15 @@ def load_reference_clusters(searchtype: str) -> Dict[str, ReferenceCluster]:
     return load_reference_clusters_from_dir(data_dir)
 
 
-def load_reference_proteins(searchtype: str) -> Dict[str, Protein]:
-    """ Load protein database
+def load_reference_proteins_from_dir(data_dir: str) -> dict[str, Protein]:
+    """ Loads a protein database from the given directory.
 
         Arguments:
-            searchtype: determines which database to use, allowable values:
-                            clusterblast, subclusterblast, knownclusterblast
-        Returns:
-            a dictionary mapping protein name to Protein instance
-    """
-    options = get_config()
-    if searchtype == "clusterblast":
-        logging.info("ClusterBlast: Loading gene cluster database proteins into memory...")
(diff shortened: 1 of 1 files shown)


SJShaw:
Updated with changes to ClusterBlast to allow for easier reuse of database preparation and loading.

PR #848, opened 2025-08-05
Small(ish) bug fixes

SJShaw:
Changes the handling of CDS features where both start coordinate and end coordinate were marked as ambiguous (e.g. those features found/predicted with protein homology). In 8.0.1 these would occasionally crash when the feature had introns and contained a domain and the region genbank attempted to generate nucleotide coordinates for a domain feature. In 8.0.2 the handling was changed to abort when attempting to extract protein location from a doubly ambiguous CDS. Unfortunately, features like these are reasonably common in RefSeq annotations, so this has now been changed so that affected features:
- have translations removed when the translation is longer than the location
- have translations generated when missing (including when removed above), then setting the start coordinate to unambiguous
- will only be handled when converting from Genbank/biopython features, directly creating a `CDSFeature` with a doubly ambiguous location will be rejected

Fixes a number of other, smaller bugs:
- when at least 2 domains from an equivalence group were present in a gene, all domains in the gene were treated as if they were in the group, leading to some infrequent cases of cluster loss
- the `FAAL_cd05931` profile introduced with the `polyyne` rule was too generous, and while this didn't cause extra clusters to be called, it did cause more issues with the bug above and had many false positives recorded in gene functions
- an edge case when a GFF file contained a circular record *and* a CDS was annotated with an end coordinate at the end of the record, that gene was split into cross-origin sections with the latter part being `[0:0]`, triggering an assertion that the start wasn't less than the end
- the gene table in HTML output had a hardcoded size in pixels, so browser zooming and higher resolutions wasted a great deal of space

Lastly, while catching some typing issues making these changes, a newer `mypy` found some useful issues. The pinned version has been updated to that newer version and existing typing issues fixed.

### DIFF
--- a/antismash/common/gff_parser.py
+++ b/antismash/common/gff_parser.py
@@ -186,7 +186,7 @@ def split_cross_origin_locations(features: list[SeqFeature], length: int) -> Non
         # most circular records won't have introns, but handle it just in case
         parts = []
         for part in feature.location.parts:
-            if part.end < length:
+            if part.end <= length:
                 # don't change the location, but convert to secmet for better usability
                 parts.append(FeatureLocation(part.start, part.end, part.strand))
             elif part.start >= length:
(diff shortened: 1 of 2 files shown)


kblin (approved):
Looks good to me 👍

PR #851, opened 2025-08-15
nrps_pks: fix invalid colinear ordering when reversing cross-CDS modules

SJShaw:
Colinear orderfinding of NRPS/PKS candidate had issues where CDS features with cross-CDS modules were in the forward strand and there was also a CDS in the reverse strand with a module containing a terminal domain.

The previous logic would flip the order of all genes when that reverse strand terminal was found, which then inverted the order of those chains of CDS features with cross-CDS modules, leading to some very wrong results.

For example, here is CP102510 from 70,000 to 14,0000, with the reverse strand PKS block containing a terminal and the right hand NRPS side being a chain of three CDS features joined by cross-CDS modules:
image

With `JIW86_40515` and `JIW86_40520` being represented twice, each with only some of their domains.

image

After the fix, it's now correct:
image

This particular example is almost certainly two separate products, but it serves well as a test case.

### DIFF
--- a/antismash/modules/nrps_pks/orderfinder.py
+++ b/antismash/modules/nrps_pks/orderfinder.py
@@ -464,17 +464,46 @@ def find_colinear_order(cds_features: List[CDSFeature]) -> List[CDSFeature]:
         Returns:
             a list of CDSFeatures in estimated order
     """
+    if not cds_features:
+        return []
+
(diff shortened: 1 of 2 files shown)


kblin (approved):
Looks good to me 👍

PR #887, opened 2026-03-17
More detailed output file settings, as used by the antiSMASH database

kblin:


kblin:
superseded by #903 and #904

PR #888, opened 2026-03-18
config: Allow loading config defaults from environment variables

kblin:
This fixes #885


kblin:
After some more discussion, this will need to be redone completely to only allow overriding the database location and the CPU counts to allow for the use case in #885 without potentially breaking too many other things

PR #890, opened 2026-03-24
Fixes for missing/invalid main database directory

SJShaw:
A number of modules and sub-modules crash out instead of nicely reporting issues when a database directory was missing or had missing contents. Other issues were also present, where attempts to prepare some data are made despite the base data not being present.


kblin:
Other than the extra warning, this looks fine

kblin (approved):
Looks good to me 👍

kblin (approved):
Looks good to me 👍

PR #896, opened 2026-04-08
sideloader: fix HTML for cross-origin regions listing sideloaded areas for the whole record

SJShaw:
This bug was introduced with circular genome support and the bounds checks for areas to include in the HTML details panel didn't correctly use circular-aware methods.

The PR also extends the `DummyRegion` test helper to allow passing args to any automatically constructed child areas.


kblin (approved):
Looks good to me 👍

PR #905, opened 2026-06-11
Declutter HTML overview

SJShaw:
Currently there is a field in the region table for the product/category of the best MIBiG match. Some users are overwhelmed by it, some ignore it, and some use it to help determine whether the match is relevant. To help those first two groups, the field can simply be removed. To also account for the latter, the "low" confidence matches will now be hidden in the table when the MIBiG top-level product categories aren't included in the region's set of cluster type categories. The exception here is the MIBiG category `other`, a kind of catch-all, which will always be considered as included for the type checks.

The detail sections within each region page are unchanged.

Using `NZ_AMPN02000002` as an example, before:
image

And after, with 4.9 no longer showing a match due to being low confidence with mismatching types:
image


kblin (approved):
Looks good to me 👍

PR #906, opened 2026-06-11
Update the deazapurine detection rule

josecediel25:
## Summary

This PR updates the QueE HMM profile and refines the deazapurine detection rule.

These changes are motivated by our recent large-scale computational identification of >900 deazapurine-containing BGCs across bacteria (https://doi.org/10.64898/2026.04.15.718813), which provide a significantly expanded sequence space for pHMM model building compared to the previous training set (~32 sequences).

---

## Motivation

The previous QueE HMM profile was constructed from a limited set of sequences (~32), which does not capture the full diversity observed in newly identified deazapurine BGCs.

Additionally, the current detection logic is overly restrictive and misses deazapurines derived from CDG.

---

### Updated QueE HMM profile
- Rebuilt using a diverse dataset of >900 QueE sequences from newly identified deazapurine-containing BGCs

### Updated deazapurine detection rule

**Previous rule:**
- GTP_cyclohydroI AND PTPS AND QueE AND QueC AND NOT (dpdA1 OR bTGT)

**Updated rule:**
- GTP_cyclohydroI AND PTPS AND (QueE OR QueC) AND NOT (dpdA1 OR bTGT)

This modification allows detection of BGCs in which deazapurines can originate from either CDG or preQ0 as building blocks, without requiring the formation of preQ0 (i.e., cases where QueC may be absent or may serve a tailoring role).

---

## Biological Rationale

The updated rule captures additional CDG-derived metabolites, including:

- Tubercidin (https://doi.org/10.1186/s12934-018-0978-8)
- Roseomycin (https://doi.org/10.1021/jacs.5c13285)

These examples demonstrate that requiring both QueE and QueC underestimates the diversity of deazapurine biosynthesis

---

## Data Availability

The sequences used to construct the updated QueE HMM profile are available at:

https://doi.org/10.5281/zenodo.20647444

---

Best,  
José


kblin:
Do you happen to have an example (as in GenBank ID) that the old rule didn't catch but the new rule does catch?

josecediel25:
> Do you happen to have an example (as in GenBank ID) that the old rule didn't catch but the new rule does catch?

Yes, tubercidin: MG706975.1

kblin:
I've added a corresponding `EXAMPLE` line to the rules, and updated the PR on top of the current codebase, it should be good to go once the tests pass

kblin (approved):
Looks good to me 👍

PR #908, opened 2026-06-26
Incremental update for library support

SJShaw:
This a subset of #847, containing the changes with least complications and improving on some of them.

PR #912, opened 2026-07-28
Misc small fixes

SJShaw:
This PR includes a few fixes that only required a few line changes:

HTML generation now strips more irrelevant whitespace in the output, both during and after rendering. For S. Coelicolor on my machine, this reduced the runtime of the full HTML module by ~5% and reduced the output file size by ~3%. I don't expect browser load/render time will be noticeably improved, as most of the current time is in rendering the content rather than parsing the file itself.

The terpene module used `--minimal` in the full runs of integration testing, which didn't test the module's HTML generation. This caused `cds_selector_span()`'s change in #908 to no be accounted for and caused crashes on any input with a terpene cluster. The template was updated and the integration tests were adjusted to now include HTML generation.

CDS features with `codon_start` qualifiers are adjusted during runtime to be 0-indexed, but weren't adjusted back to 1-indexed prior to writing output files, resulting in chained use of those output files to have incorrect values. Those adjustments are now correctly undone prior to output.

Also includes a fix for a missing f-string in the rule parser.

### DIFF
--- a/antismash/common/hmm_rule_parser/rule_parser.py
+++ b/antismash/common/hmm_rule_parser/rule_parser.py
@@ -756,7 +756,7 @@ class SingleCondition(Conditions):
 
     def get_hit_string(self) -> str:
         if self.negated:
-            return "{self.hits}*({self.name})"
+            return f"{self.hits}*({self.name})"
         return f"{self.hits}*{self.name}"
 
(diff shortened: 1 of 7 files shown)


kblin (approved):
Looks good to me 👍

PR #913, opened 2026-07-29
chore: update PR linting action for compatibility

SJShaw:
Github changed `pull_request_target` in a way that caused the existing action to no longer function. The fix is to change to `pull_request`, but it's a good time to change the permissions to be even more restrictive than they were. The action has now been split into two jobs with separate permission sets.

The first builds the lint comparison report with minimal access/capabilities, notably removing the pull-request write permissions that it previously had.

The second updates the PR with the result of the comparison, with elevated pull-request write permissions in order for it to modify the PR (e.g. make the comment about the changes in linting messages).

### DIFF
--- a/.github/workflows/pylint_issue_reporter.yml
+++ b/.github/workflows/pylint_issue_reporter.yml
@@ -1,13 +1,12 @@
 name: Comment on pylint issue comparison
 
-on: [pull_request_target]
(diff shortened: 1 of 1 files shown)


kblin (approved):
Looks good to me 👍

PR #914, opened 2026-07-30
chore: split PR linting into multiple workflows for external fork support

SJShaw:
My previous version of this in #913 didn't properly account for permissions with external forks. Combined with a misunderstanding of how action permissions worked when that action was changed within in a PR, it snuck through into main.

As opposed to internal branches, external forks don't allow elevating to write permissions at all (with `pull_request`), so separating into two jobs was insufficient. Now those jobs have become two separate _workflows_, which allows the second workflow to have elevated permissions to make the lint comment on PRs. The comment workflow doesn't have the context for which PR is involved, so that is passed through in the action artifact from the linter workflow.

### DIFF
--- a/.github/workflows/pr_commentor.yml
+++ b/.github/workflows/pr_commentor.yml
@@ -1,0 +1,42 @@
+name: Comment on PR
+
+on:
+  workflow_run:
+    workflows: ["Lint comparison"]
+    types:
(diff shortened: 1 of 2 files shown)


kblin (approved):
Looks good to me 👍

PR #916, opened 2026-08-24
Use vanilla CSS instead of SASS

kblin:
This conversion gets rid of the libsass dependency by converting to modern vanilla CSS. This should work in any browser released in the last three years.

Draft PR just to keep track of this, needs a small change in `antismash-js` to also add the `secmet` helper class to be fully functional.


kblin:
ok, I guess the unwanted file check requires this to be structured differently

PR #918, opened 2026-08-24
Drop SASS dependency, use modern vanilla CSS instead

kblin:


kblin:
ok, that clearly didn't work. So I guess this PR depends on #917 being merged first so the unwanted file check doesn't fail

kblin:
Drop the libsass dependency, use plain CSS instead. This should work on any browser released in the last three years.

Before merging, antismash-js needs to add support for the new helper CSS class first.

PR #919, opened 2026-08-28
Remove the CASSIS module

SJShaw:
The CASSIS module was deprecated in the last major release. This PR removes it. While there's nothing left that uses `meme` from the MEME suite, that section of `antismash.common.subprocessing` is untouched and can be removed as a whole when FIMO is also no longer needed.


kblin (approved):
Looks good to me 👍

